PR #3, opened 2020-07-27
Concrete implementation

liamsi:
Rendered Readme:
https://github.com/lazyledger/nmt/blob/ismail/add_concrete_implementation/README.md

TODOs: 
 - [x] more tests (ideally > 90 % test-coverage)
 - [x] ~~decide between returning error or bool for verification methods (currently both)~~ verification should just be a predicate  
 - [x] add verifiaction equivalent to `Prove(idx)`
 - [x] clean up verification
 - [x] add proper Readme 
 - [x] improve godoc documentation

closes https://github.com/lazyledger/nmt/issues/4


nmt.go
```diff
@@ -0,0 +1,142 @@
+package nmt
+
+import (
+	"bytes"
+	"crypto"
+	"errors"
+	"fmt"
+
+	"github.com/liamsi/merkletree"
```
musalbas, line 9: This repo should be moved to lazyledger org

nmt.go
```diff
@@ -0,0 +1,142 @@
...
+func (n NamespacedMerkleTree) NamespaceSize() int {
+	return n.nidLen
+}
+
+func (n NamespacedMerkleTree) Push(data NamespacePrefixedData) error {
+	got, want := data.NamespaceSize(), n.NamespaceSize()
+	if got != want {
+		return fmt.Errorf("%w: got: %v, want: %v", ErrMismatchedNamespaceSize, got, want)
+	}
+	// TODO: or should we only push to the actual tree at the end
```
musalbas, line 39: In the LazyLedger prototype, the code for block production is responsible for making sure the data is pushed in order.

However, as a standalone library, it probably makes sense to handle this here and make it idiotproof.

nmt.go
```diff
@@ -0,0 +1,142 @@
...
+	// when we compute the root? The we can push messages bundled by namespaces together: e.g. s.t.
+	// they are lexicographically ordered by namespace IDs
+	// (first transactions,then intermediate state roots, evidence, messages etc)
```
musalbas, line 42: In the LazyLedger prototype, the code for block production is responsible for making sure the data is pushed in order.

However, as a standalone library, it probably makes sense to handle this here and make it idiotproof.

nmt.go
```diff
@@ -0,0 +1,142 @@
...
+//nolint:errcheck
+func (n NamespacedMerkleTree) CompactRoot() []byte {
+	h := n.baseHasher.New()
+	minNs, maxNs, root := n.Root()
+	h.Write(minNs)
+	h.Write(maxNs)
+	h.Write(root)
+	// TODO: should we merkelize these 3 values instead?
+	// Or add in some form of domain seperation here too?
+	return h.Sum(nil)
```
musalbas, line 64: In what circumstances would a hash of the root with its namespaces prefix be necessary? Sending the actual root with the prefixes doesn't take up much more bandwidth.

liamsi: @adlerjohn argued that the library should rather have a method to return a compact commitment to of the prefixed root (e.g. via `hash(minNs||maxNs||actualRoot)`).

nmt.go
liamsi: It's a fork of the NebolousLabs' tree with very few changes. Ideally, we can merge these changes into their library. If that doesn't happen, I'll move it to the LL org 👍  
(see: https://gitlab.com/NebulousLabs/merkletree/-/issues/21)

musalbas: There are other changes that we are likely going to need to make to the underlying merkletree implementation that may not be mergeable upstream. For example, we should probably change the implementation to store key=>value pairs of nodes in the tree using the same interface as the SMT MapStore, so that we can plug in other MapStores like an IPFS-based one, and reuse the same code: https://github.com/lazyledger/smt-ipfs-mapstore

This is something that needs to be investigated a bit though, as from my understanding the NebulousLabs merkletree library I think may compute the root on demand and doesn't keep around any of the old nodes in memory. Alternatively, it may be possible to implement this MapStore storage functionality within the NMT library, which uses the MapStore as "cache", and only invokes the NebulousLabs merkletree library to recompute roots if the node data is unavailable in the MapStore.

nmt.go
liamsi: removed in https://github.com/lazyledger/nmt/pull/3/commits/fd6c90fa94c346a153d7f9537a4cb64a70273670 for now as the caller can handle this as required

nmt.go
liamsi: > This is something that needs to be investigated a bit though, as from my understanding the NebulousLabs merkletree library I think may compute the root on demand and doesn't keep around any of the old nodes in memory.

That's my understanding too. An alternative could be to look further into trillian, it allows to recompute the tree quickly [in memory](https://github.com/google/trillian/blob/7166f485ebc965d24116db667c9e058a8ed9275c/merkle/memory_merkle_tree.go#L69) (I did some measurements and it seemed a bit faster then NebolousLabs') but also comes with a storage backend where you can query for ranges etc: https://github.com/google/trillian/blob/master/storage/log_storage.go

My plan for now was to abstract storage away and give the prover/verifier some kind of leaf iterator.

liamsi:
Besides the prover API this should be complete. @adlerjohn @musalbas you can give it another pass if you like (I'll refrain from force pushing to this branch from now on).

nmt.go
```diff
@@ -0,0 +1,268 @@
...
+		}
+	}
+	proofMessages := n.leafs[proofStart:proofEnd]
+	if found {
+		return proofStart, proofEnd, proof, proofMessages, nil
+	}
+	// XXX further clarify the edge-case (!found && !foundRangeStart)
+	// !found: we return leafHashes instead:
+	return proofStart, proofEnd, proof, nil, n.leafHashes[proofStart:proofEnd]
+}
```
liamsi, line 111: Something I wondered about: If the namespace to prove `nID` can't be found in the tree and `nID < minNS` or `nID > maxNS` (where `minNs, maxNs,_ := tree.Root()`), shouldn't the prove of exclusion be either an inclusion proof of the left- or right-most leaf?

Currently, in the [prototype code](This should be the same logic as https://github.com/lazyledger/lazyledger-prototype/blob/2aeca6f55ad389b9d68034a0a7038f80a8d2982e/simpleblock.go#L105-L160
) as well as above application proofs return:
`(proofStart, proofEnd, proof, messages, hashes) == (0, 0, nil, nil, nil)` if `nID < minNS || maxNS < nID`.

iface.go
```diff
@@ -1,23 +1,35 @@
...
+	// Should return an error if the namespace ID size of the input
+	// does not match the tree's NamespaceSize().
+	Push(data NamespacePrefixedData) error
 
 	// Return the namespaced Merkle Tree's root together with the
 	// min. and max. namespace ID.
 	Root() (minNs, maxNS NamespaceID, root []byte)
 }
+
+type NamespacedProver interface {
```
musalbas, line 17: Why is this in a separate interface and not part of the tree interface?

liamsi: Anything wrong with that? I prefer to keep interfaces as small as possible and it seems to be good practice: Golang is a bit extreme as many interfaces in the standard library are one-method ifaces only (Reader, Writer, Formatter, CloseNotifier and many more). As a type can implement several interfaces there is no reason to describe all behaviour in one interface. 

I could imagine a optimized version of the tree that does take in a bunch of data and returns the root as quickly as possible. Or, a simplified version of the NMT that is just used to generate diagrams. 
Or, a prover that fetches its data differently than the standard implementation and independently from a NMT instance (e.g. from the network). 

If you feel strongly about this, I can change it.

nmt.go
liamsi: Capturing the discussing with @adlerjohn and @musalbas:
 - in the case the that  min and max namespace of the root is not within range of the nid, no proof is required
 - the root already signifies that this nid does not exist in that particular tree
 - retuning a sentinel error (signifying and as before returning no proof) probably makes the most sense here

iface.go
musalbas: Ah so is NamespacedMerkleTree also a NamespacedProver? So as long as developers don't have to create a separate instance of NamespacedMerkleTree and NamespacedProver to make proofs, that seems OK.

liamsi: No they do not need to create separate instances. It's just wherever in the LL-core code we would expect a NMT we can pass in anything that implements the NMT interface, and if we need a prover, anything that implements the Prover interface. This is only half true though: a prover and verifier actually belongs to a particular tree implementation. There is no real way to communicate this via the go type system (e.g. there is no equivalent to [associated types](https://www.microsoft.com/en-us/research/wp-content/uploads/2005/01/assoc.pdf)).

Note that it is not clear if we even need these interfaces at all! It was just a good way to quickly communicate the public API without having to look scroll through implementation details. I guess keeping them only make sense if we want this library to have contain several different implementations (with the same interface). It might actually be a good idea to remove the interfaces before merging and re-introduce them when we actually see a real need for them...

liamsi: removed the interfaces for now

nmt.go
liamsi: Push now errors if the namespaces pushed aren't pushed "in order".

liamsi:
TODOs: 
 - [ ] more tests (ideally > 90 % test-coverage)
 - [ ] clean up verification
 - [ ] add proper Readme and improve documentation

nmt.go
```diff
@@ -0,0 +1,242 @@
...
+		if prevNs.Less(nID) && nID.Less(currentNs) {
+			if !foundRangeStart {
+				foundRangeStart = true
+				proofStart = index
+			}
+			proofEnd = index + 1
+		}
+		prevLeaf = curLeaf
+	}
+	return proofStart, proofEnd
```
liamsi, line 158: Doesn't the lexicographic sorting imply that `proofStart == proofEnd-1` is always true? 

Meaning that a proof of absence will always be just *one* leaf hash: namely at the single index i for which `n_{i-1} < nID < n_i`; if that was true for another index `j`:  `n_{j-1} < nID < n_j` that would imply for instance n_i < n_j but also n_j < n_i (which is impossible and hence proves that there can't be another index). 

Anything wrong with the reasoning here?

musalbas: Yes, this is the case in the prototype: https://github.com/lazyledger/lazyledger-prototype/blob/master/simpleblock.go#L133

liamsi: That's not obvious: in the code you share `proofStart` gets set to `index` once but `proofEnd` could be set to `index + 1` multiple times (or at least it is not obvious that it only gets set once), and index is increasing (hence proofEnd could be anything > proofStart). I think it will only be set once if the data is pushed sorted by namespace (which is enforced here); so we can safely assume one leaf.

proof.go
```diff
@@ -118,6 +119,9 @@ func (proof Proof) VerifyNamespace(nth Hasher, nID namespace.ID, data []namespac
 			gotLeafHashes = append(gotLeafHashes, hashLeafFunc(gotLeaf.Bytes()))
 		}
 	}
+	if !proof.IsOfAbsence() && len(gotLeafHashes) != (proof.End()-proof.Start()) {
+		return false, ErrMissingData
+	}
```
liamsi, line 124: without this, the verification might panic (e.g. when removing a leaf before calling `VerifyNamespace`.

liamsi:
Only thing left to do here is: https://github.com/lazyledger/nmt/pull/3#discussion_r461415836

README.md
```diff
@@ -1,16 +1,89 @@
...
+  namespace identifiers of the messages.
+  In a namespaced Merkle tree, each non-leaf node in
+  the tree contains the lowest and highest namespace
+  identifiers found in all the leaf nodes that are descendants of the non-leaf node, in addition to the hash of
+  the concatenation of the children of the node. This
+  enables Merkle inclusion proofs to be created that prove to a verifier that all the elements of the tree for
+  a specific namespace have been included in a Merkle
+  inclusion proof.
+
+The concept was first introduced by [@musalbas] in the LazyLedger [academic paper].
```
musalbas, line 23: Missing link?

README.md
```diff
@@ -1,16 +1,89 @@
...
+nidSize := uint8(1)
+// the leaves that will be pushed
+data := []namespace.PrefixedData{
+    namespace.PrefixedDataFrom(namespace.ID{0}, []byte("leaf_0")),
+    namespace.PrefixedDataFrom(namespace.ID{0}, []byte("leaf_1")),
+    namespace.PrefixedDataFrom(namespace.ID{1}, []byte("leaf_2")),
+    namespace.PrefixedDataFrom(namespace.ID{1}, []byte("leaf_3"))}
+// use the hasher to set the namespace size as well as 
+// the underlying hash function:
+nmtHasher := defaulthasher.New(nidSize, crypto.SHA256)
```
musalbas, line 37: Shouldn't it be defaultHasher for camel case consistency?

musalbas: Ah this is another package, nvm.

musalbas: Doesn't this need to be imported first in the example?

defaulthasher/hasher_test.go
```diff
@@ -0,0 +1,107 @@
+package defaulthasher
```
musalbas, line 1: It would be a good idea to add a bulk test like https://github.com/lazyledger/smt/blob/optimised_compute/bulk_test.go for building the tree and testing proofs. This can pick-up unexpected edge cases.

nmt.go
```diff
@@ -0,0 +1,240 @@
...
+	namespaceRanges map[string]merkletree.LeafRange
+	minNID          namespace.ID
+	maxNID          namespace.ID
+}
+
+func New(treeHasher Hasher) *NamespacedMerkleTree {
+	return &NamespacedMerkleTree{
+		treeHasher: treeHasher,
+		tree:       merkletree.NewFromTreehasher(treeHasher),
+		// XXX: 100 seems like a good capacity for the leaves slice
```
musalbas, line 38: What are the consequences of this parameter? Does it mean that only up to 100 leafs can be pushed?

musalbas:
I think a bulk test should be added like https://github.com/lazyledger/smt/blob/optimised_compute/bulk_test.go for building the tree and testing proofs. This can pick-up unexpected edge cases, otherwise only the base cases are tested.

README.md
liamsi: All URLs for links are at the at the bottom of the page. Makes it easier to read the raw text. Also you have a good overview of all references at one place. Links look and work the same though (see the rendered version: https://github.com/lazyledger/nmt/blob/ismail/add_concrete_implementation/README.md)

README.md
liamsi: I wanted to keep the example minimal and not include the usual boilerplate code (imports + `func main()`). Can change that if you think this is confusing.

nmt.go
liamsi: This only determines the initial capacity. Meaning up to 100 items should be appendable without an additional mem allocation. `append`s after that initial capacity will increase the underlying capacity s.t. the append will still work (the capacity will roughly be doubled as needed: https://github.com/golang/go/blob/ba9e10889976025ee1d027db6b1cad383ec56de8/src/runtime/slice.go#L125).

liamsi:
Thanks for the feedback. Good point, will add in something like this 👍

nmt.go
adlerjohn: Given that the number of shares will always be a power of 2 for LazyLedger, would it not be better to set this initial capacity to a power of 2 as well? An original square with size 64*64 (extended 128*128) would give ~1MB blocks, so 128 here seems like a reasonable parameter choice for initial capacity to avoid having to always do a re-size operation.

liamsi: Later, maybe we should expose this as a param to the constructor. e.g. in LL the caller will know the exact capacity and if that's passed in, only one allocation will be necessary. This should not matter if the allocations happen on the stack only. I never fully understood which stuff ends up on the heap vs stack in golang but apparently this allocation is on the heap (which is more expensive). Check for yourself via: `go build -gcflags -m`
> [...]
> ` make([][]byte, 0, 100) escapes to heap`
> [...]

liamsi: > so 128 here seems like a reasonable parameter choice for initial capacity 

I agree.

README.md
liamsi: ok, I'll add in the full `package main+`import`+`func main`.

```go
package main

import (
    "crypto"
    _ "crypto/sha256"

    "github.com/lazyledger/nmt/defaulthasher"
    "github.com/lazyledger/nmt/namespace"
)

func main() {
    // the tree will use this namespace size
    nidSize := uint8(1)
    // the leaves that will be pushed
    data := []namespace.PrefixedData{
        namespace.PrefixedDataFrom(namespace.ID{0}, []byte("leaf_0")),
        namespace.PrefixedDataFrom(namespace.ID{0}, []byte("leaf_1")),
        namespace.PrefixedDataFrom(namespace.ID{1}, []byte("leaf_2")),
        namespace.PrefixedDataFrom(namespace.ID{1}, []byte("leaf_3"))}
    // use the hasher to set the namespace size as well as 
    // the underlying hash function:
    nmtHasher := defaulthasher.New(nidSize, crypto.SHA256)
    tree := nmt.New(nmtHasher)
    for _, d := range data {
        tree.Push(d)
    }
    
    // compute the root
    root := tree.Root()
    // the root's min/max namespace is the min and max namespace of all leaves:
    if root.Min().Equal(namespace.ID{0}) {
        fmt.Printf("Min namespace: %x\n", root.Min())
    }
    if root.Max().Equal(namespace.ID{1}) {
        fmt.Printf("Max namespace: %x\n", root.Max())
    }
    
    // compute proof for namespace 0:
    proof, err := tree.ProveNamespace(namespace.ID{0}); if err != nil {
        panic("unexpected err")
    }
    
    // verify proof using the root and the leaves of namespace 0:
    leafs := []namespace.PrefixedData{
        namespace.PrefixedDataFrom(namespace.ID{0}, []byte("leaf_0")),
        namespace.PrefixedDataFrom(namespace.ID{0}, []byte("leaf_1"))}
    
    if proof.VerifyNamespace(nmtHasher, namespace.ID{0}, leafs, root) {
        fmt.Printf("Successfully verified namespace: %x\n", namespace.ID{0})
    }
    
    if proof.VerifyNamespace(nmtHasher, namespace.ID{2}, leafs, root) {
        panic(fmt.Sprintf("Proof for namespace %x, passed for namespace: %x\n", namespace.ID{0}, namespace.ID{2}))
    }
}
```

liamsi: resolved in https://github.com/lazyledger/nmt/pull/3/commits/129dba0cfe9b1e090b49714fc334537a69ece018

nmt.go
liamsi: Changed to 128 in https://github.com/lazyledger/nmt/pull/3/commits/056112da486a306f9e88ac60f306025a9cebb9ed

We might want to consider the following options (potentially in followup PRs):
 - make this a param in the constructor
   - pros: full control / can be optimized by use-case (small vs big trees)
   - cons: leaky abstraction? (leaks the internals of the implementation in some sense)
 - make the constructor take in all the leafs at once :
   - pros: optimal with regards to allocs & does not leak any internals; might be easier to read as you can pass in the whole data at once
   - cons: ?

musalbas: Perhaps there can be a method to change this parameter, instead of putting in the constructor and leaking the abstraction?

namespace/id.go
```diff
@@ -0,0 +1,24 @@
+package namespace
+
+import "bytes"
+
+type ID []byte
```
adlerjohn, line 5: Is there a way to make this a fixed-size byte array? I know Go doesn't have generics, but it's a little opaque to have the ID length only set at runtime when it's a parameter that can be known at compile time. Maybe using constants?

defaulthasher/hasher.go
```diff
@@ -0,0 +1,95 @@
...
+func (n *DefaultHasher) HashNode(l, r []byte) []byte {
+	h := n.New()
+	// the actual hash result of the children got extended (or flagged) by their
+	// children's minNs || maxNs; hence the flagLen = 2 * NamespaceLen:
+	flagLen := 2 * n.NamespaceLen
+	leftMinNs, leftMaxNs := l[:n.NamespaceLen], l[n.NamespaceLen:flagLen]
+	rightMinNs, rightMaxNs := r[:n.NamespaceLen], r[n.NamespaceLen:flagLen]
+
+	minNs := min(leftMinNs, rightMinNs)
+	maxNs := max(leftMaxNs, rightMaxNs)
```
adlerjohn, line 68: This doesn't seem to implement https://github.com/lazyledger/lazyledger-specs/pull/59

namespace/data.go
```diff
@@ -0,0 +1,36 @@
+package namespace
+
+type PrefixedData struct {
+	namespaceLen uint8
```
adlerjohn, line 4: Is the type of namespace ID going to be abstracted to a type alias?

defaulthasher/hasher.go
liamsi: It doesn't yet. I want to implement that part in a separate PR.

namespace/data.go
liamsi: Like this? https://github.com/lazyledger/nmt/blob/056112da486a306f9e88ac60f306025a9cebb9ed/namespace/id.go#L5

Or do you mean as an interface (which also could make sense)?

adlerjohn: In this case, for length, maybe `type NamespaceLen uint8`? That way you're not passing around raw `uint8`s without context.

namespace/id.go
liamsi: I don't think there is an elegant way to achieve this in golang but still allow the user to define the size they like. An alternative would be to make the ID length/size part of the "prefixed data" type (I miss associated types ...), which then could return an ID (of fixed length) depending on the concrete implementation (similar to https://golang.org/pkg/crypto/sha256/#Sum256).

namespace/digest.go
```diff
@@ -0,0 +1,54 @@
...
+
+func NewIntervalDigest(min, max ID, digest []byte) IntervalDigest {
+	return IntervalDigest{
+		min:    min,
+		max:    max,
+		digest: digest,
+	}
+}
+
+func IntervalDigestFromBytes(nIDLen uint8, digestBytes []byte) IntervalDigest {
```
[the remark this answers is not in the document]
liamsi: Good point!

treehasher.go
```diff
@@ -0,0 +1,24 @@
...
+	// HashLeaf defines how a leaf is hashed.
+	HashLeaf(leaf []byte) []byte
+
+	// HashNode defines how a inner node is hashed.
+	HashNode(l, r []byte) []byte
+
+	// Size returns the size of the underlying hasher.
+	Size() int
+	// Return the size of the namespace
+	NamespaceSize() uint8
```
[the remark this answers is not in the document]
liamsi: Hmm, it's just an interface and the Size depends on the concrete implementation.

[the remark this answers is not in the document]
adlerjohn: Okay, then you should document the implementation with what specifically the size is!

namespace/id.go
adlerjohn: Ugh 🤷. My concern is that variable-sized arrays might serialize differently than fixed-size arrays, depending on which canonical serialization scheme is used. And this library should be parametrizable with different serialization schemes.

liamsi: > My concern is that variable-sized arrays might serialize differently than fixed-size arrays, depending on which canonical serialization scheme is used.

The length will be fixed for any tree even with the current implementation. The type doesn't reflect this to be flexible. Similar to the `Sum` method in https://golang.org/pkg/hash/#Hash which also always return the same size for a concrete implementation of a hash function.

> And this library should be parametrizable with different serialization schemes.

Hmm, good point. I'll think about how to abstract this cleanly. Is it OK, if I document this better for this PR and think about this in future iterations?

namespace/data.go
liamsi: > That way you're not passing around raw uint8s without context.

That is true, but the param name also communicates some context. Also, unless the type alias doesn't also have Uint8 in its name, it won't communicate the min and may values here ([0, 255]). This reminds me, that I need to disallow `0` everywhere.

liamsi:
I actually want to add in some more changes (besides adding "bulk" tests) before merging:
 - [ ] hide away/remove the `Hasher` abstraction and only let the user choose a base hasher (and a namespace size in bytes)
 - [ ] introduce/extract an interface `namespace.Data` that abstracts away how the internal bytes are laid out (currently just bytes prefixed with the namespace ID) 
 - [ ] on that interface rename `Bytes()` to `MarshalBinary`  (maybe embedding this: https://golang.org/pkg/encoding/#BinaryMarshaler)

namespace/data.go
liamsi: Introduced a type alias in the namespace package: `type Size uint8`. So when using it, it will read: `namespace.Size`. You can still use `uint8` and `namespace.Size` more or less interchangeably...
regarding the 0 namespace size: this should just correspond to a regular (non-namespaced) merkle tree. Will add tests for that case too.

namespace/id.go
liamsi: > And this library should be parametrizable with different serialization schemes.

Thinking about this further: I don't think the NMT should care about serialization. The whole API and the implementation would be simpler if the caller would need to pass in the data already properly serialized (as bytes / simple type aliases).

Also, I think going back to my original suggestion to have `Push()` take in the pair of namespace and data separately instead of some wrapping type makes more sense.

adlerjohn: So one problem is that hash digests are also a data type and also need a canonical serialization. This can't be done at the input level because hashing is an internal operation to the Merkle tree.

liamsi: By the hash digests you mean the resulting root, right? These are also just bytes. Can't the consumer serialize these as he wishes? Isn't this a bit like saying the sha256 `hash.Hash`  additionally should define a canonical serialization format for its output :thinking: E.g. the SMT doesn't define how to pass the resulting root bytes over the wire. Why would it be different here?

Can you clarify the expectations the point you are trying to make a bit further?

treehasher.go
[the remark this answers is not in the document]
liamsi: This only lives in the internal package which does not belon to the public API (nor can it be used by any other library). The Size is propoerly documented in anything public facing.

namespace/digest.go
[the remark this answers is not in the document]
liamsi: https://github.com/lazyledger/nmt/pull/3/commits/46b99b2138428e094506abea4527ec21c2ecdcb0

PR #10, opened 2020-09-07
Optionally ignore max namespace ID

### LINKED ISSUE #8: Add the possibility to "ignore" namespace IDs
In other words implement: https://github.com/lazyledger/lazyledger-specs/pull/59

liamsi:
implements and closes #8 

Also tackles setting the init. capacity for internal slices as an option (see comment/discussion: https://github.com/lazyledger/nmt/pull/3#discussion_r467980031)

### DIFF
--- a/nmt.go
+++ b/nmt.go
@@ -84,14 +84,15 @@ type NamespacedMerkleTree struct {
 func New(h hash.Hash, setters ...Option) *NamespacedMerkleTree {
 	// default options:
 	opts := &Options{
-		InitialCapacity: 128,
-		NamespaceIDSize: 32,
+		InitialCapacity:    128,
+		NamespaceIDSize:    32,
+		IgnoreMaxNamespace: false,
 	}
 
 	for _, setter := range setters {
 		setter(opts)
 	}
-	treeHasher := internal.NewNmtHasher(opts.NamespaceIDSize, h)
+	treeHasher := internal.NewNmtHasher(h, opts.NamespaceIDSize, opts.IgnoreMaxNamespace)
 	return &NamespacedMerkleTree{
 		treeHasher:      treeHasher,
 		tree:            merkletree.NewFromTreehasher(treeHasher),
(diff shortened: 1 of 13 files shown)


liamsi:
needs more testing (moving back to draft)

nmt.go
```diff
@@ -41,36 +81,42 @@ type NamespacedMerkleTree struct {
...
 // If the namespace size is 0 this corresponds to a regular non-namespaced
 // Merkle tree.
-func New(h hash.Hash, namespaceSize namespace.Size) *NamespacedMerkleTree {
-	treeHasher := internal.NewNmtHasher(namespaceSize, h)
+func New(h hash.Hash, setters ...Option) *NamespacedMerkleTree {
+	// default options:
+	opts := &Options{
+		InitialCapacity:    128,
+		NamespaceIDSize:    32,
+		IgnoreMaxNamespace: false,
```
liamsi, line 89: Should we default `IgnoreMaxNamespace` to true? The only user will us for some time and we expect to use it with ignoring the parity share ID.

Also: did we settle on the 32 bytes for a namespace ID in the spec? IMO 8 or 16 bytes should be enough.
```suggestion
		NamespaceIDSize:    16,
		IgnoreMaxNamespace: true,
```

adlerjohn: I don't see any issue with defaulting to `true`, since the NMT is supposed to be used in conjunction with data availability proofs anyways.

For the namespace ID size, nothing is final, but you can open an issue in https://github.com/LazyLedger/lazyledger-specs to discuss tradeoffs with different sizes.

liamsi: OK, I'll set `IgnoreMaxNamespace` to true for now and leave the 32 bytes for now (as this seems to be the closest to the spec rn, e.g. given the constants and the reserved namespace IDs). I'll open an issue tomorrow :-)

liamsi: https://github.com/lazyledger/lazyledger-specs/issues/62

musalbas:
It would be a good idea to add some tests to ensure that the max and min namespace IDs are what we expect them to be when max namespace ID is ignored.

liamsi:
There already is [TestIgnoreMaxNamespace](https://github.com/lazyledger/nmt/blob/cd3fc3c7261912ad86a0cde90e634604709ff0d9/nmt_test.go#L292) but I agree that we should extend this and at least add these cases: 
 - only leafs with max namespace IDs are pushed (2, 3, ... leafs)
 - 2 leafs: one with max namespace the other without
 - more than 4 leafs & half the leafs with max namespace

liamsi:
https://github.com/lazyledger/nmt/pull/10/commits/b85998ba0d8ce42f7083ed1292351f8c7683ccfb

nmt_test.go
```diff
@@ -291,63 +291,176 @@ func TestNamespacedMerkleTree_ProveNamespace_Ranges_And_Verify(t *testing.T) {
...
+		{"single leaf with MaxNID (ignored)",
+			true,
+			[]namespace.Data{namespace.PrefixedData8(append(maxNID, []byte("leaf_1")...))},
+			maxNID,
+		},
+		{"single leaf with MaxNID (not ignored)",
+			false,
+			[]namespace.Data{namespace.PrefixedData8(append(maxNID, []byte("leaf_1")...))},
+			maxNID,
+		},
```
liamsi, line 317: The single leaf case is special in the sense as the only namespace won't be ignored. (We will have powers of 2 in LL anyways though).

liamsi:
Merging as is. If there are any objections, let's follow up in separate PRs/issues.

PR #14, opened 2020-12-25
Simplify computing the tree's root and improve perf

liamsi:
This refactoring:

 - reduces the time to compute the tree root: ~40% faster - see below 
 - uses a simpler to grasp, recursive method to compute the tree root (instead of using NebolousLabs' implementation)
 - this also makes it easier to store or cache the inner tree nodes (we might not need to actually store the inner nodes; will summarize this in a separate issue)
 - removes the `namespace.Data` interface 
 - `Push`'s signature changed to `Push(id namespace.ID, data []byte) error` instead of `Push(data namespace.Data) error`

A followup PR could further remove the dependency on NebolousLabs' tree completely (or the dep on our [fork](https://github.com/lazyledger/merkletree)).

As for the benchmarks, I've added a simple benchmark for 64, 128, and 256 leaves. I've added the benchmark on the current master and on this branch. Before and after:

```
$ benchstat before.txt after.txt 
name                       old time/op    new time/op    delta
ComputeRoot/64-leaves-12      182µs ± 2%     114µs ± 2%  -37.14%  (p=0.000 n=30+30)
ComputeRoot/128-leaves-12     363µs ± 2%     228µs ± 2%  -37.20%  (p=0.000 n=28+29)
ComputeRoot/256-leaves-12     728µs ± 2%     456µs ± 1%  -37.40%  (p=0.000 n=28+27)

name                       old alloc/op   new alloc/op   delta
ComputeRoot/64-leaves-12     46.5kB ± 0%    52.1kB ± 0%  +12.26%  (p=0.000 n=30+30)
ComputeRoot/128-leaves-12    87.3kB ± 0%    98.1kB ± 0%  +12.39%  (p=0.000 n=29+29)
ComputeRoot/256-leaves-12     180kB ± 0%     196kB ± 0%   +9.22%  (p=0.000 n=30+30)

name                       old allocs/op  new allocs/op  delta
ComputeRoot/64-leaves-12      1.04k ± 0%     0.65k ± 0%  -37.30%  (p=0.000 n=30+30)
ComputeRoot/128-leaves-12     2.07k ± 0%     1.30k ± 0%  -37.34%  (p=0.000 n=30+30)
ComputeRoot/256-leaves-12     4.13k ± 0%     2.58k ± 0%  -37.39%  (p=0.000 n=30+30)
```

Note the distinct memory allocations occurred per op (single iteration) and the time per iteration reduced by ~40%. The number of bytes allocated per op increased slightly (9-12%) though. This is negligible compared to the other improvements (and can likely be improved further later on).

If you want to reproduce the benchmarks, apply this patch on master / the commit before this PR is merged:
<details>
  <summary>Click to expand diff</summary>
  
```patch
Index: nmt.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/nmt.go b/nmt.go
--- a/nmt.go	(revision db7185d82ed479bc478eaaa8c6d9c22981318a10)
+++ b/nmt.go	(date 1608897797194)
@@ -84,7 +84,7 @@
 	// default options:
 	opts := &Options{
 		InitialCapacity:    128,
-		NamespaceIDSize:    32,
+		NamespaceIDSize:    8,
 		IgnoreMaxNamespace: true,
 	}
 
Index: nmt_test.go
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/nmt_test.go b/nmt_test.go
--- a/nmt_test.go	(revision db7185d82ed479bc478eaaa8c6d9c22981318a10)
+++ b/nmt_test.go	(date 1608897755523)
@@ -7,7 +7,9 @@
 	"encoding/binary"
 	"fmt"
 	"math"
+	"math/rand"
 	"reflect"
+	"sort"
 	"testing"
 
 	"github.com/lazyledger/nmt/namespace"
@@ -542,6 +544,38 @@
 	})
 }
 
+func BenchmarkComputeRoot(b *testing.B) {
+	b.ReportAllocs()
+	tests := []struct {
+		name      string
+		numLeaves int
+		nidSize   int
+		dataSize  int
+	}{
+		{"64-leaves", 64, 8, 256},
+		{"128-leaves", 128, 8, 256},
+		{"256-leaves", 256, 8, 256},
+	}
+
+	for _, tt := range tests {
+		data := generateRandNamespacedRawData(tt.numLeaves, tt.nidSize, tt.dataSize)
+		b.ResetTimer()
+		b.Run(tt.name, func(b *testing.B) {
+			for i := 0; i < b.N; i++ {
+				n := New(sha256.New())
+				for j := 0; j < tt.numLeaves; j++ {
+					if err := n.Push(namespace.PrefixedData8(data[j])); err != nil {
+						b.Errorf("err: %v", err)
+					}
+				}
+				_ = n.Root()
+			}
+		})
+	}
+
+}
+
+
 func shouldPanic(t *testing.T, f func()) {
 	//nolint:errcheck
 	defer func() { recover() }()
@@ -599,3 +633,24 @@
 
 	return h.Sum(nil)
 }
+
+func generateRandNamespacedRawData(total int, nidSize int, leafSize int) [][]byte {
+	data := make([][]byte, total)
+	for i := 0; i < total; i++ {
+		nid := make([]byte, nidSize)
+		rand.Read(nid)
+		data[i] = nid
+	}
+	sortByteArrays(data)
+	for i := 0; i < total; i++ {
+		d := make([]byte, leafSize)
+		rand.Read(d)
+		data[i] = append(data[i], d...)
+	}
+
+	return data
+}
+
+func sortByteArrays(src [][]byte) {
+	sort.Slice(src, func(i, j int) bool { return bytes.Compare(src[i], src[j]) < 0 })
+}

```
</details>


liamsi:
Need to update the readme before R4R.

nmt.go
```diff
@@ -64,14 +65,16 @@ func IgnoreMaxNamespace(ignore bool) Option {
```
evan-forbes, line 65: Github won't let me comment on line 46 because it wasn't changed, but it should be updated from 
```go
```

liamsi: Thanks! Updated in 88ca665

adlerjohn:
Is there a reason the benchmark code isn't part of the PR's changes?

liamsi:
What do you mean? It is part of this PR:
https://github.com/lazyledger/nmt/blob/88ca665fec7d624841ff1b320fda8e357f6883d4/nmt_test.go#L549-L578

It isn't part of the master branch and that's what the results of the entry comment compare to. To reproduce the comparison, you need to apply the patch/diff from the entry comment on master and either manually compare results or use [benchstat](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat).

adlerjohn:
> It is part of this PR:

Oh I see. I interpreted

> If you want to reproduce the benchmarks, apply this patch on master / the commit before this PR is merged

As 

> If you want to reproduce the benchmarks, apply this patch on the commit before this PR is merged

I.e. it's not part of the PR proper.

liamsi:
Oh, I see. Yeah, that sentence can be confusing. Sorry about that.

adlerjohn (changes requested):

adlerjohn (approved):

Issue #9, opened 2020-09-07
Add possibility to store nodes of the tree (MapStore)


liamsi:
Also reuse stored hashes instead of recomputing them each time. 

Copying the related [review comment](https://github.com/lazyledger/nmt/pull/3#discussion_r461555647) here:
> There are other changes that we may need to make to the underlying merkletree implementation that may not be mergeable upstream. For example, we should probably change the implementation to store key=>value pairs of nodes in the tree using the same interface as the SMT MapStore, so that we can plug in other MapStores like an IPFS-based one, and reuse the same code: https://github.com/lazyledger/smt-ipfs-mapstore

> This is something that needs to be investigated a bit though, as from my understanding the NebulousLabs merkletree library I think may compute the root on demand and doesn't keep around any of the old nodes in memory. Alternatively, it may be possible to implement this MapStore storage functionality within the NMT library, which uses the MapStore as "cache", and only invokes the NebulousLabs merkletree library to recompute roots if the node data is unavailable in the MapStore.

related to: #6

liamsi:
#16 is sufficient to store the nodes on IPFS. Will close this one for now. There is another issue about storing the inner nodes in memory here: #17.

Issue #15, opened 2021-01-08
Remove NebulousLabs' merkletree dependency for proofs as well


liamsi:
#14 removed the dependency on NebulousLabs' implementation for computing the root of the tree and replaced it a very simple recursive algo that boils down to a few lines of code (derived from the recursive definition of [rfc6962](https://tools.ietf.org/html/rfc6962#section-2.1) and very similar to the simple merkle tree implementation in [tendermint](https://github.com/tendermint/tendermint/blob/03a6fb277702bf6d7773d6519e23fed1f1933930/crypto/merkle/tree.go#L9-L21). 

It would be nice if the same internal mechanism would be used for generating/verifiying proofs (instead of using the range proofs of nebolousLabs).

PR #16, opened 2021-01-13
Add possibility to traverse inner nodes

liamsi:
Adds an _optional_ "NodeVisitor" function to the tree that will be called on each node of the tree.

This can be used to tackle #9, and can also be used in the IPLD-plugin. It might also be useful to inspect/debug the tree, or, to print-out the tree in different ways.

### DIFF
--- a/nmt.go
+++ b/nmt.go
@@ -21,12 +21,16 @@ const (
 var (
 	ErrMismatchedNamespaceSize = errors.New("mismatching namespace sizes")
 	ErrInvalidPushOrder        = errors.New("pushed data has to be lexicographically ordered by namespace IDs")
+	noOp                       = func(hash []byte, children ...[]byte) {}
 )
+
+type NodeVisitorFn = func(hash []byte, children ...[]byte)
 
 type Options struct {
 	InitialCapacity    int
 	NamespaceIDSize    namespace.IDSize
 	IgnoreMaxNamespace bool
+	NodeVisitor        NodeVisitorFn
 }
 
 type Option func(*Options)
@@ -63,8 +67,15 @@ func IgnoreMaxNamespace(ignore bool) Option {
 	}
 }
 
+func NodeVisitor(nodeVisitorFn NodeVisitorFn) Option {
+	return func(opts *Options) {
+		opts.NodeVisitor = nodeVisitorFn
+	}
+}
+
 type NamespacedMerkleTree struct {
 	treeHasher internal.NmtHasher
+	visit      NodeVisitorFn
 
 	// just cache stuff until we pass in a store and keep all nodes in there
 	// currently, only leaves and leafHashes are stored:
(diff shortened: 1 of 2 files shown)


evan-forbes:
This is so dope!

liamsi:
Looks like this is all we need for the IPLD plugin.

musalbas:
As `NodeVisitorFn` is stateless (it just takes in the hash of a node and its children), how would the function determine which nodes are part of which Merkle proof or is it not relevant?

liamsi:
Good point. Currently, this is not relevant - at least for the first LL ipld plugin it isn't and I tried to keep the diff as small as possible. That said, there probably are use-cases where this will become relevant (e.g. when doing #15 this might need to change) 🤔 Did you have something specific in mind when you wondered about this? Maybe I'm missing something?

For now, I don't see any urgent need: nodes who generate proofs are expected to have all the data to generate these, and, they can always recompute the (inner) proof-nodes from scratch. The `NodeVisitorFn` is only called when computing the root.

musalbas:
I assume this is used in the IPLD plugin in order to publish the intermediate nodes? If so this seems like a non-generalisable feature but I guess it's OK if it's implemented as an option rather than a mandatory constructor argument.

musalbas (approved):

liamsi:
Yes, that is exactly what it is used for. What kind of generalization do you have in mind though? For reading the nodes from storage? 

If there is sth. obvious which do need soon, we can add this here too. Otherwise, let's track this in a separate issue.

evan-forbes (approved):

musalbas:
> Yes, that is exactly what it is used for. What kind of generalization do you have in mind though? For reading the nodes from storage?
> 
> If there is sth. obvious which do need soon, we can add this here too. Otherwise, let's track this in a separate issue.

I suppose actually storing the tree itself (in memory) within NMT itself and letting people read from it.

liamsi:
The tree exists in memory but proofs are currently recomputed on demand. That's isn't "reading from it" but should be quite efficient already, especially for a first version / mvp (the leaf-data and the leaf-hashes are also stored in memory and the former can be retrieved via the API).

If we want to avoid recomputing the inner proof nodes every time they are needed, we can actually store the inner nodes too. Not sure this would need to be exposed as an exported API or sth like this. Of course, if we want to store and then read the inner nodes from somewhere else, we'd need to change the API. We could also pass in an optional `NodeGetterFn` or `SubTreeGetter` maybe? For now I think recomputing the nodes is fine and will be fast enough, especially way faster than retrieving them from any kind of persistent storage (especially much faster thanreading them from sth like IPFS). Does that make sense?

adlerjohn:
Hashing is pretty fast and involves no disk I/O. Storing inner nodes would mean multiple on-disk accesses along with more storage overhead. So it might not actually be so bad to just fetch the leaves all at once then recompute proofs as needed.

Issue #18, opened 2021-02-09
Export nmt-specific hashers


liamsi:
It might make sense to export the nmt-specific hashing currently "hidden" in the internal package:
https://github.com/lazyledger/nmt/tree/master/internal

This particularly makes sense in the light of the IPLD plugin. Instead of redefining the logic when adding a custom hasher (see: https://github.com/lazyledger/lazyledger-core/pull/155, https://github.com/lazyledger/go-multihash/pull/1 and in particular [this (wip) code](https://github.com/lazyledger/go-multihash/blob/88bad1265973e3d986433b60817d2e49a72812f3/sum.go#L184-L242)), we could just import the hashers from here instead. Besides exporting it would require some minor changes to the hasher api to easily match that of [HashFunc](https://github.com/multiformats/go-multihash/blob/6f1ea18f1da5f7735ea31b5e2011da61c409e37f/sum.go#L28).

The main benefit of doing this is that the logic is defined in one place only. As a result, we don't need to keep two implementations in sync. Furthermore,  we do not need to make sure (e.g. via unit tests) that the hasher used in the nmt internally matches the one used to compute the Sum of an IPLD block (as the registered custom multihasher just imports the code from here).

Note to use the hashers as described above, ideally, we would register two multihashers instead of one (and looking at the prefix as we do [currently](https://github.com/lazyledger/go-multihash/blob/88bad1265973e3d986433b60817d2e49a72812f3/sum.go#L184-L242)).

PR #19, opened 2021-02-12
Restructure code to expose the internal hasher:

liamsi:
In lazyledger/lazyledger-core#155 we want to register a custom multihash that matches those used by the nmt internally. Instead of duplicating the code for this (as in https://github.com/lazyledger/go-multihash/pull/1), we simply export the API to construct an nmt hasher. 

 - expose NmtHasher API s.t. it can be used to construct and use NMT Hashsers externally (e.g. for IPLD multihashers)
 - Add two functions `Sha256Namespace8FlaggedLeaf`, `Sha256Namespace8FlaggedInner` whose signature matches regular hash functions that simplify the above even further

TODOs:

 - [x] explain the need for this change in the context of https://github.com/lazyledger/lazyledger-core/pull/155, https://github.com/lazyledger/go-multihash/pull/1
 - [x] add tests for `Sha256Namespace8FlaggedLeaf`, `Sha256Namespace8FlaggedInner`

closes: #18


hasher.go
```diff
@@ -0,0 +1,156 @@
...
+// If the input does not fulfil this, we will panic.
+// The output will also be of length 2*DefaultNamespaceIDLen+sha256.Size = 48 bytes.
+func Sha256Namespace8FlaggedInner(leftRight []byte) []byte {
+	const flagLen = DefaultNamespaceIDLen * 2
+	sha256Len := defaultHasher.Size()
+	left := leftRight[flagLen:+sha256Len]
+	right := leftRight[flagLen+sha256Len:]
+
+	return defaultHasher.HashNode(left, right)
+}
```
liamsi, line 51: Adding these 2 methods is the only real change. Everything else is moving existing code from the internal package to the main nmt one, and, deleting some no longer needed code/abstractions.

hasher.go
```diff
@@ -0,0 +1,159 @@
...
+
+func (n *Hasher) IsMaxNamespaceIDIgnored() bool {
+	return n.ignoreMaxNs
+}
+
+func (n *Hasher) NamespaceSize() namespace.IDSize {
+	return n.NamespaceLen
+}
+
+func NewNmtHasher(baseHasher hash.Hash, nidLen namespace.IDSize, ignoreMaxNamespace bool) *Hasher {
```
tzdybal, line 72: As this method is in `nmt` package, it can be named `NewHasher`.

tzdybal (approved):
LGTM

evan-forbes (approved):

PR #22, opened 2021-03-12
Remove panic from IntervalDigestFromBytes

### LINKED ISSUE #20: Work around the panic thrown by IntervalDigestFromBytes
[IntervalDigestFromBytes](https://github.com/lazyledger/nmt/blob/master/namespace/digest.go#L14) is meant to deserialize bytes commonly received from the wire that does not guarantee to compel with the panic check, allowing a non-valid network message to kill an application that uses the lib. We can either:
1. Throw an error instead
2. Force lib user to do an additional check before or recover from panic :)

But obviously, the first one is more preferable.

Wondertan:
Fixes #20

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -183,7 +183,7 @@ func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID na
 	if verifyCompleteness {
 		// leftSubtrees contains the subtree roots upto [0, r.Start)
 		for _, subtree := range leftSubtrees {
-			leftSubTreeMax := namespace.IntervalDigestFromBytes(nth.NamespaceSize(), subtree).Max()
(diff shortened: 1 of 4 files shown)


proof.go
```diff
@@ -220,3 +220,13 @@ func nextSubtreeSize(start, end uint64) int {
 	}
 	return 1 << uint(ideal)
 }
+
+// mustIntervalDigestFromBytes optimistially converts bytes to IntervalDigest or panics
+func mustIntervalDigestFromBytes(idlen namespace.IDSize, bytes []byte) namespace.IntervalDigest {
```
Wondertan, line 225: Don't think we need to export this

liamsi (approved):
Thanks @Wondertan 🙏🏼

Wondertan:
@liamsi, ok will do that

Issue #24, opened 2021-03-18
Create a rsmt2d.Tree interface wrapper for NamespacedMerkleTree


evan-forbes:
We need a wrapper that fulfills the `rsmt2d.Tree` interface. That wrapper will need to be able to create nmt roots assuming that the namespaces are included in the data being pushed, but only for the first half of the tree. This means the wrapper should be aware of the square size ahead of time. 

[realted ll-core issue](https://github.com/lazyledger/lazyledger-core/issues/172)

liamsi:
That wrapper doesn't necessarily need to live in this repo, right? It could be in ll-core for instance.

> This means the wrapper should be aware of the square size ahead of time.

Another way to achieve the same could be to let ll-core prefix everything with namespace ID, even the parity bytes. Then the tree can differentiate between real shares and (second half) parity shares because the latter would carry the max. namespace.

evan-forbes:
I don't think it will be difficult to move the logic over to lazyledger-core, and it makes a lot of sense to keep all lazyledger-core logic in lazyledger-core, so I'll close this issue along with the PR.

PR #25, opened 2021-03-19
Implement wrapper for rsmt2d.Tree

evan-forbes:
This PR creates a wrapper for `NamespacedMerkleTree` that conforms to the `rsmt2d.Tree` interface, which will enable [`RetrieveBlockData`](https://github.com/lazyledger/lazyledger-core/pull/232) to properly repair the extended data square. It will also help close [#172](https://github.com/lazyledger/lazyledger-core/issues/172). 

ref [#15](https://github.com/lazyledger/rsmt2d/issues/15)

closes: #24


wrapper_test.go
```diff
@@ -0,0 +1,46 @@
...
+
+	"github.com/lazyledger/rsmt2d"
+)
+
+const (
+	shareSize           = 256
+	adjustedMessageSize = shareSize - DefaultNamespaceIDLen
+)
+
+func TestPushErasuredNamespacedMerkleTree(t *testing.T) {
```
evan-forbes, line 14: I wasn't sure what else I could test, so any suggestions are greatly appreciated.

go.mod
```diff
@@ -5,4 +5,5 @@ go 1.14
 require (
 	github.com/google/gofuzz v1.2.0
 	github.com/lazyledger/merkletree v0.0.0-20201214195110-6901c4c3c75f
+	github.com/lazyledger/rsmt2d v0.0.0-20210121155026-23bf69f65287
```
liamsi, line 8: IMO, the NMT should not need to know about rsmt2d. That is why I think the wrapper should live in ll-core. There all the dependencies need to come together anyways. And introducing the dependency here might cause some headaches in the future.

wrapper.go
```diff
@@ -0,0 +1,88 @@
+package nmt
+
+import (
+	"crypto/sha256"
+
+	"github.com/lazyledger/nmt/namespace"
+	"github.com/lazyledger/rsmt2d"
+)
+
+var paritySharesNamespaceID = namespace.ID{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF}
```
liamsi, line 10: Note that the tree already knows about this internally: https://github.com/lazyledger/nmt/blob/e0a317aa2d0ff84fcf1478527e4c90e6a13245d7/hasher.go#L77

wrapper.go
```diff
@@ -0,0 +1,88 @@
...
+		panic("tree size exceeded")
+
+	// if the namespace is included in the data, use that ns
+	case w.pushCount+1 <= w.squareSize/2:
+		copy(nsID, data[:DefaultNamespaceIDLen])
+
+	// if the data is erasure data use the parity ns
+	default:
+		copy(nsID, paritySharesNamespaceID)
+	}
```
liamsi, line 61: Couldn't the caller do this instead? They already know which shares are parity shares. Then they could simply prefix the parity shares with `namespace.ID{0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF}`. It would simplify the wrapper a bit. 

On another note, I think we should change the method signature for `Push` from `Push(nid, data []byte)` (currently) to `Push(data []byte)` for the underlying nmt anyways. I think I've mentioned this in some convos but I never got around opening a small issue explaining the rationale for that 😬 Will do in a sec.

wrapper_test.go
liamsi: I think there is not much to test as the wrapper is really simple. And that is good. Well done!

The only thing I'd suggest is to also cover the non-happy paths in unit tests (e.g. the few edge cases where this can panic/error).

liamsi (commented):
IMO, the same can be achieved without any changes to the NMT. Namely by adding the wrapper to ll-core. Otherwise, the PR looks good to me. I want to think further about the implications of introducing rsmt2d as a dependency here before approving.

go.mod
adlerjohn: So what's the plan here, is the logic in this PR going to be moved instead of lazyledger-core?

evan-forbes:
I'm going to close this PR for now. As mentioned in the comments, it makes sense to keep any new logic  in lazyledger-core. I think we will still need a wrapper, because we need to pass a `rsmt2d.Tree` to `rsmt.RepairExtendedDataSquare` when recovering the data. That tree also needs to be aware of our hard coded parity namespace for the second half of the tree.

wrapper.go
evan-forbes: +1 for changing the signature

We will need the wrapper to be aware of the parity share, as it will have to be passed to `rsmt2d.RepairExtendedDataSquare`

PR #27, opened 2021-03-19
Change Push signature and simplify PrefixedData

liamsi:

PR #28, opened 2021-03-19
Implement Equal for IntervalDigest

Wondertan:
Otherwise, a user has to implement it himself using Bytes() what would cause him to make unnecessary allocations

### DIFF
--- a/namespace/digest.go
+++ b/namespace/digest.go
@@ -1,6 +1,9 @@
 package namespace
 
-import "fmt"
+import (
+	"bytes"
+	"fmt"
+)
 
 type IntervalDigest struct {
 	min    ID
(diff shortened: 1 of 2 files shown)


Wondertan:
(hope the fact that I didn't create an issue for this won't hurt)

namespace/digest.go
```diff
@@ -24,7 +27,7 @@ func IntervalDigestFromBytes(nIDLen IDSize, digestBytes []byte) (IntervalDigest,
 	}, nil
 }
 
-func (d IntervalDigest) Min() ID {
+func (d IntervalDigest) Min() ID {// TODO: I don't like that we cannot compare Digests without allocating to much
```
liamsi, line 30: Either remove the todo or turn into sth more constructive. What is needed to reduce allocations?
Also, ideally, they would start with `TODO(Wondertan):`

liamsi (commented):
Thanks, looks good! Can you edit that TODO? Let's merge then. 

I'll create a release after we've merged in #27 too.

namespace/digest.go
Wondertan: Haha, that's an accidental paste from another piece of work. It does not even make sense anymore. Will force push it

Wondertan:
@liamsi, TODO is now removed

liamsi (changes requested):
One last thing, can you make sure the code is `go fmt`d? The linter is complaining about the file `namespace/digest_test.go`.

Wondertan:
@liamsi, done. For some reason that did not happen automatically

liamsi (approved):
👍

Issue #26, opened 2021-03-19
reconsider Push to take in the pair (nid, data) as one argument instead


liamsi:
`Push`  currently looks like this:
https://github.com/lazyledger/nmt/blob/e0a317aa2d0ff84fcf1478527e4c90e6a13245d7/nmt.go#L249

The reason for this, is that in a few parts of the code, we need to operate on the namespaceID independently. Also, at the very beginning of the implementation, I did not fully understand that having a one to one mapping of one NMT <-> one namespace size would be the only reasonable use-case. The implementation already stores the namespace size and internally could split and merge the passed in argument into (nid, data) and back.

The only downside is that the API looks a bit less clear. This can easily be mitigated by having a well documented type alias `type NamespacedData []byte`.

I'm proposing to change Push to ` func (n *NamespacedMerkleTree) Push(ndata NamespacedData) error`.

This would also slightly simplify the wrapper that is needed for the rsmt2d library (see #25 ).

PR #29, opened 2021-03-22
Fix minor refactoring glitch introduced with #27

liamsi:
This fixes another minor glitch that was introduced by changing the Push method signature in #27. Similar to what @evan-forbes found in https://github.com/lazyledger/nmt/pull/27#discussion_r598746721

Followup to / part of #26.


evan-forbes:
Whoops, we might have missed another one at https://github.com/lazyledger/nmt/blob/2ce94e57fee491ef48a324222ce2149c3a2b7314/nmt_test.go#L489 

just triple checked, so that should be the last of them.

liamsi:
Mehr, I thought I've gone through each call of `Push`but it seems I've missed the one you've found. Thanks again, Evan!

Wondertan (approved):

PR #30, opened 2021-03-29
Fix nasty append bug that caused 'min' namespace id modifications 

Wondertan:
https://github.com/lazyledger/lazyledger-core/pull/248#issuecomment-809595268

### DIFF
--- a/namespace/digest.go
+++ b/namespace/digest.go
@@ -40,7 +40,11 @@ func (d IntervalDigest) Hash() []byte {
 }
 
 func (d IntervalDigest) Bytes() []byte {
-	return append(append(d.min, d.max...), d.digest...)
+	return append(append(append(
+		make([]byte, 0, len(d.min)*2+len(d.digest)),
+		d.min...),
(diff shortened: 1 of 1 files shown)


namespace/digest.go
```diff
@@ -40,7 +40,11 @@ func (d IntervalDigest) Hash() []byte {
 }
 
 func (d IntervalDigest) Bytes() []byte {
-	return append(append(d.min, d.max...), d.digest...)
+	return append(append(append(
+		make([]byte, 0, len(d.min)*2+len(d.digest)),
+		d.min...),
+		d.max...),
+		d.digest...)
```
liamsi, line 47: @adlerjohn will love this one 🤣 

Note that differently from the title, the nid does not get modified. Still the race detector would raise an alarm here before as this is seen as a WRITE.

liamsi (approved):
👏🏼 Thanks! Did you verify that this actually fixes the problem with the race detector in ll-core?

PR #31, opened 2021-04-07
Use accelerated SHA-256 computatiion library

Wondertan:
This is a simple PR that swaps native SHA-256 calculation lib to [ASM accelerated one](https://github.com/minio/sha256-simd) that should give decent performance improvement.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,4 +5,5 @@ go 1.14
 require (
 	github.com/google/gofuzz v1.2.0
 	github.com/lazyledger/merkletree v0.0.0-20201214195110-6901c4c3c75f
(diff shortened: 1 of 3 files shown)


Wondertan:
If the proposal is ok, I will create further PRs in other repos that uses SHA-256

adlerjohn:
This is sick!

liamsi:
I think @musalbas did some benchmarks a few years ago. Similarly, there is this https://github.com/tendermint/tendermint/issues/2186

Wondertan:
Closing as per sync discussion: we would like to keep `nmt` dependency-free, while a user can still use `sha256-simd` by providing custom hasher to `nmt`

PR #32, opened 2021-05-03
nmt_test: add err message for panic

renaynay:


liamsi (approved):
Makes sense and LGMT. Thanks a lot for your contribution @renaynay 🙏👌
There might be one or two other places like this in the code I think.

evan-forbes (approved):
dope! :+1: 
thumbs-up

PR #33, opened 2021-05-03
.gitignore: add gitignore file

renaynay:


.gitignore
```diff
@@ -0,0 +1 @@
+.idea
```
liamsi, line 1: Can you add a comment, that is for JetBrains IDEs (goland etc).

While at it, we could also typical go-related ignores as well. I think these 2 make sense for instance:

https://github.com/github/gitignore/blob/218a941be92679ce67d0484547e3e142b2f5f6f0/Go.gitignore#L9

https://github.com/github/gitignore/blob/218a941be92679ce67d0484547e3e142b2f5f6f0/Go.gitignore#L12

liamsi (approved):
👍LGTM

PR #35, opened 2021-05-10
Make `IntervalDigest` jsonable

### LINKED ISSUE #34: Make `IntervalDigest` json encodable
Some [tests](https://github.com/lazyledger/lazyledger-core/blob/59ceaf8c44a069a55d5f8f8f3bea50b23a186146/rpc/client/evidence_test.go#L115) in lazyledger-core will require `DataAvailabilityHeader` to be encoded via json. 

This issue needs to be solved before [#312](https://github.com/lazyledger/lazyledger-core/pull/312) can be merged.

The json representation of IntervalDigest might be useful for #21 as well.

evan-forbes:
Some [tests](https://github.com/lazyledger/lazyledger-core/blob/59ceaf8c44a069a55d5f8f8f3bea50b23a186146/rpc/client/evidence_test.go#L115) in lazyledger-core require `DataAvailabilityHeader` to be encoded via json, this PR exports the fields of IntervalDigest, along with adding appropriate json tags.

This PR, or another alternative, should be merged before [#312](https://github.com/lazyledger/lazyledger-core/pull/312) can be merged.

closes #34

### DIFF
--- a/namespace/digest.go
+++ b/namespace/digest.go
@@ -6,9 +6,9 @@ import (
 )
 
 type IntervalDigest struct {
-	min    ID
-	max    ID
-	digest []byte
+	Min    ID     `json:"max"`
+	Max    ID     `json:"min"`
+	Digest []byte `json:"digest"`
 }
 
 // IntervalDigestFromBytes is the inverse function to IntervalDigest.Bytes().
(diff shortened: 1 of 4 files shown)


namespace/digest.go
```diff
@@ -6,9 +6,9 @@ import (
 )
 
 type IntervalDigest struct {
-	min    ID
-	max    ID
-	digest []byte
+	Min    ID     `json:"max"`
+	Max    ID     `json:"min"`
```
liamsi, line 10:
```suggestion
	Min    ID     `json:"min"`
	Max    ID     `json:"max"`
```

evan-forbes: omg :eyes: whoops

namespace/digest.go
```diff
@@ -6,9 +6,9 @@ import (
 )
 
 type IntervalDigest struct {
-	min    ID
-	max    ID
-	digest []byte
+	Min    ID     `json:"min"`
+	Max    ID     `json:"max"`
+	Digest []byte `json:"digest"`
 }
```
liamsi, line 12: I'm wondering if we should not get rid of this struct entirely and instead store this as raw bytes min||max||digest instead. For instance, when using this with ipfs, the CID will essentially be this plus some overhead bytes that stem from the multihash standard. 
We can merge this as is to unblock https://github.com/lazyledger/lazyledger-core/pull/312 but we should track this somewhere.

liamsi (approved):
👍🏼

evan-forbes:
will merge after [#317](https://github.com/lazyledger/lazyledger-core/pull/317) CI passes, then I'll update that PR to use nmt v0.3.2

PR #38, opened 2021-05-13
(bugfix): create a new hash instance each time for go-routine safety

### LINKED ISSUE #37: bug: Hasher is not go-routine safe
On lazyledger-core, we access NMTs hasher (e.g when hashing leaves) in several go routines. 

These will try to re-use the same hasher instance: https://github.com/lazyledger/nmt/blob/6dc97cc6f9fdd1c800108d2cab06a4d8b088ea8a/hasher.go#L98

We should create a new instance instead. 

This might be related to: https://github.com/lazyledger/lazyledger-core/issues/330

At least tests with larger blocks panic with:

```
panic: d.nx != 0

goroutine 85 [running]:
crypto/sha256.(*digest).checkSum(0xc00043d9f0, 0x0, 0x0, 0x0, 0x0)
        c:/go/src/crypto/sha256/sha256.go:234 +0x1e0
crypto/sha256.(*digest).Sum(0xc00009eb00, 0x0, 0x0, 0x0, 0xc0, 0xb6, 0x0)
        c:/go/src/crypto/sha256/sha256.go:210 +0x70
```

liamsi:
- [x]  add test to reproduce the error here (without https://github.com/lazyledger/lazyledger-core/commit/152202137455450f971df5c7294380714dcc7ac6)

closes #37

### DIFF
--- a/hasher.go
+++ b/hasher.go
@@ -15,9 +15,11 @@ const (
 	DefaultNamespaceIDLen = 8
 )
 
+type NewHashFn = func() hash.Hash
+
 // defaultHasher uses sha256 as a base-hasher, 8 bytes
 // for the namespace IDs and ignores the maximum possible namespace.
-var defaultHasher = NewNmtHasher(sha256.New(), DefaultNamespaceIDLen, true)
+var defaultHasher = NewNmtHasher(sha256.New, DefaultNamespaceIDLen, true)
 
 // Sha256Namespace8FlaggedLeaf uses sha256 as a base-hasher, 8 bytes
 // for the namespace IDs and ignores the maximum possible namespace.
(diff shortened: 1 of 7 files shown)


hasher.go
```diff
@@ -46,19 +48,26 @@ func Sha256Namespace8FlaggedLeaf(namespacedData []byte) []byte {
...
-	hash.Hash
+	newHashFn    func() hash.Hash
 	NamespaceLen namespace.IDSize
 
 	ignoreMaxNs      bool
 	precomputedMaxNs namespace.ID
+	size             int
+}
+
+func (n *Hasher) Size() int {
```
Wondertan, line 66: Why is this needed? Hasher(Hash).Size() was called somewhere before? From the codecov bot it seems like this is not called anywhere.

liamsi: It's implictly used by the [`TreeHasher`](https://github.com/lazyledger/nmt/blob/8cd1c48591ee5b84075b1f968628005796f974c6/internal/subtree_hasher.go#L9) which is currently still used by various methods `Prove`,  `ProveNamespace` and maybe more.

nmt_test.go
```diff
@@ -615,6 +616,28 @@ func BenchmarkComputeRoot(b *testing.B) {
 
 }
 
+func Test_Root_RaceCondition(t *testing.T) {
```
liamsi, line 619: This test fails on current master.

Wondertan (approved):
Besides the question of why `Fn` at the end of `NewHashFn`, have not other comments

liamsi:
Updated the name in https://github.com/lazyledger/nmt/pull/38/commits/8472fbee16034e760c6b6f5e23b539add6ea2616

PR #40, opened 2021-05-23
Reuse hash 

liamsi:
After reviewing https://github.com/lazyledger/lazyledger-core/pull/351 I've realized that #38 introduced an unnecessary alloc per leaf and inner node. 

This PR reverts the changes of #38, caches the root s.t. it only gets computed when necessary but keeps that test that aims to reproduce the issue in #37. (draft as I want to confirm this really does not re-introduce the issue in test withing ll-core, too https://github.com/lazyledger/lazyledger-core/commit/152202137455450f971df5c7294380714dcc7ac6)

### DIFF
--- a/fuzz_test.go
+++ b/fuzz_test.go
@@ -30,8 +30,8 @@ func TestFuzzProveVerifyNameSpace(t *testing.T) {
 	for _, size := range testNsSizes {
 		nidDataMap, sortedKeys := makeRandDataAndSortedKeys(size, minNumberOfNamespaces, maxNumberOfNamespaces, minElementsPerNamespace, maxElementsPerNamespace, emptyNamespaceProbability)
 		t.Logf("Generated %v namespaces for size: %v ...", len(nidDataMap), size)
-		newHashFn := sha256.New
-		tree := nmt.New(newHashFn, nmt.NamespaceIDSize(int(size)))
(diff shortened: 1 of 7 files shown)


Wondertan:
It so cool that you immediately realized how to apply the change @jsign proposed here. Love this!

liamsi:
Thanks :-) We were already doing the suggested change (of https://github.com/lazyledger/lazyledger-core/pull/351) here. Only very recently (#38) we changed this for (#37). But yeah reviewing that code made me realize that our "bugfix" (quotes as the bug only happened during unit tests), wasn't the smartest one.
In general there is plenty of room for optimizations in this library (nmt). 

Two low hanging fruits are:
- have one function call that combines Push and Root (takes all leaves and immediately returns the hash)
- get rid of the nebulous labs code for proofs (remove the dependency entirely)

liamsi:
IMO, the test in #38 is probably enough confirmation that this does not re-introduce the problem pre #38/#37.

adlerjohn:
Is this branch up-to-date against `master`? It's hard to tell from the force push.

liamsi:
Yes, the force push was to rebase on master.

adlerjohn (approved):

Wondertan (approved):

PR #44, opened 2021-08-03
fix diagram 

### LINKED ISSUE #41: Min/max definitions in diagram in readme don't match specs and implementation
If I'm understanding the diagram (https://github.com/celestiaorg/nmt/blob/6d5b99ddb71ae8f8c11d7a5fedc3b84532e949a7/imgs/example_4-leaves.png) correctly, it seems to have mistakes in how the mins and maxes are calculated.  For example:

```
min_2.0 = min(min_1.0, max_1.1)
```

Judging from the specs (https://github.com/celestiaorg/celestia-specs/blob/ec98170398dfc6394423ee79b00b71038879e211/src/specs/data_structures.md) and the implementation (https://github.com/celestiaorg/nmt/blob/6d5b99ddb71ae8f8c11d7a5fedc3b84532e949a7/hasher.go#L128-L131) it should instead say:

```
min_2.0 = min(min_1.0, min_1.1)
```

And analogously for all the other non-leaf min and max definitions in the diagram.

liamsi:
fixes #41


liamsi:
rendered
image

adlerjohn (approved):

adlerjohn:
While we're there though, would it be better to use an `.svg` image?

liamsi:
Not sure, if this was to be used in a pdf, yes. But for the markdown readme this should not matter much.

PR #48, opened 2021-09-07
Remove interval digests in favor of []byte

### LINKED ISSUE #21: Define String representation of IntervalDigest
@liamsi suggested printing IntervalDigest as CID in a private discussion, however that may require additional `go-cid` dependency for the lib. Anyway, we should define a concrete representation to ensure proper test coverage in the parent repo.

### LINKED ISSUE #36: Replace the `IntervalDigest` struct with raw bytes
I'm wondering if we should not get rid of this struct entirely and instead store this as raw bytes min||max||digest instead. For instance, when using this with ipfs, the CID will essentially be this plus some overhead bytes that stem from the multihash standard.

_Originally posted by @liamsi in https://github.com/lazyledger/nmt/pull/35#discussion_r629735307_

evan-forbes:
Quick PR to stop using interval digests and only use []byte. 

pairs well with [#524](https://github.com/celestiaorg/celestia-core/pull/524), [#513](https://github.com/celestiaorg/celestia-core/pull/513)

closes #36 
closes #21


nmt.go
```diff
@@ -372,3 +372,15 @@ func (n *NamespacedMerkleTree) computeLeafHashesIfNecessary() {
...
+func MinNamespace(hash []byte, size namespace.IDSize) []byte {
+	min := make([]byte, 0, size)
+	return append(min, hash[:size]...)
+}
+
+// MaxNamespace parses the maximum namespace id from a given hash
+func MaxNamespace(hash []byte, size namespace.IDSize) []byte {
+	max := make([]byte, 0, size)
+	return append(max, hash[size:size*2]...)
+}
```
evan-forbes, line 386: we don't have to export these

nmt_test.go
```diff
@@ -154,21 +154,26 @@ func TestNamespacedMerkleTreeRoot(t *testing.T) {
...
+	totalLen := 0
+	for _, slice := range slices {
+		totalLen += len(slice)
+	}
+	out := make([]byte, 0, totalLen)
+	for _, slice := range slices {
+		out = append(out, slice...)
+	}
+	return out
+}
```
evan-forbes, line 175: I can't wait for generics in go, so we never have to write stuff like this

nmt.go
liamsi: Do you think they will ne used outside of this package? IMO they probably won't. I'd keep them private for now.

nmt.go
```diff
@@ -265,11 +265,11 @@ func (n *NamespacedMerkleTree) Push(namespacedData namespace.PrefixedData) error
 
 // Return the namespaced Merkle Tree's root together with the
 // min. and max. namespace ID.
-func (n *NamespacedMerkleTree) Root() namespace.IntervalDigest {
+func (n *NamespacedMerkleTree) Root() []byte {
```
liamsi, line 268: Maybe the godoc comment should now be more explicit that the returned bytes are min, max and the hash concatenated, i.e. `min||max||hashDigest`?

evan-forbes: good catch! resolved in https://github.com/celestiaorg/nmt/pull/48/commits/464e30105194c9653c2f7f0c6569a03bc95080f3

liamsi (approved):
👏🏼 thanks!

nusret1996:
Thanks a lot!

PR #51, opened 2021-11-10
nmt.go: export Min/MaxNamespace functions so we can use them in celestia-node

renaynay:
This PR exports the `MinNamespace` and `MaxNamespace` functions so that we can use them in the `celestia-node` repo instead of directly copying the functionality.


liamsi (approved):
LGTM.

mattdf (approved):

Issue #55, opened 2022-05-04
api: Push data with decoupled namespace


Wondertan:
Currently, `tree.Push` requires passing `namespace.PrefixedData`, which is a byte slice prepending `namespace.ID`. However, there is a use case where we don't need to push data that is not prepended with a `namespace.ID`, and it can be passed as a separate field. This is the case where we push parity data generated via rsmt2d. Currently, we prepend parity namespace to the parity shares, which is not required. This is one of the issues causing us to store an additional 8 bytes of parity namespace per parity share via NMTWrapper, as outlined in `https://github.com/celestiaorg/celestia-node/issues/183`. I propose adding an additional method like `PushWith` to the `nmt.Tree` to allow passing decoupled namespace from the share data to allow mentioned disk usage optimization.

liamsi:
This is interesting. In the past the Push method used to have these already separated:
https://github.com/celestiaorg/nmt/blob/6e8a6a5ec3fcae4015ca5dd1923ce033a0458679/nmt.go#L254

We can revert to that! Back then there was a lot of discussions around serialzation (of course there always is 🙈 ) and I think (not sure anymore) the type aliases and the API was original meant to be able to take in a type that defines its own serialzation. IMO, the tree should just take bytes anyways and serialzation is sth that should happen before and not inside the tree or anything like that.
Long story short: if we can save some storage and bandwidth with this simple change to the NMT, then let's do it! I propose to only have one push method instead an additional pushWith. Unless there is a good reason to keep both around.

liamsi:
After discussing with @Wondertan, we realized that https://github.com/celestiaorg/celestia-node/issues/183 can be tackled without modifying the NMT library at all (and instead by modifying this [visitor](https://github.com/celestiaorg/celestia-node/blob/ed4cdf03f2bd35c30d5b9874d4bc40bea6f5a061/ipld/nmt_adder.go#L34)). So we could put this on the back burner for now. 

[...]

PR #56, opened 2022-05-05
Refactoring: Push(namespace.PrefixedData)->Push(namespace.ID, []byte)

liamsi:
for details see: https://github.com/celestiaorg/nmt/issues/55 (particularly: https://github.com/celestiaorg/nmt/issues/55#issuecomment-1118931841)

closes: #55


liamsi:
Note that due to the copy: https://github.com/celestiaorg/nmt/pull/56/checks?check_run_id=6312183477
image

evan-forbes (approved):
LGTM :+1: 

everytime I come to this repo, I'm always impressed with the test coverage and benchmarks etc. @liamsi 's work a long time ago is still paying dividends

also, one day we will stop changing the API for `Push` lolol

Wondertan (commented):
@liamsi, One mid-level thing. We shouldn't pass the visitor the result of the append of nid+data and data only, or we should pass both as separate params as well.

Or just rework Visitor API completely, as we discussed 😅

liamsi:
>  or we should pass both as separate params as well.

Yeah, that is what I thought as well but after we chatted about it it seemed not necessary and I didn't think about it further. Can you open an issue with the proposed method signature? I'll do a quick follow-up PR in the next few days.

PR #57, opened 2022-05-28
Extend API to allow arbitrary leaf ranges to be proven.

liamsi:


evan-forbes (approved):
yep, that's it!

adlerjohn:
This is a backwards-compatible change, no? Why was the minor version incremented to `v0.9.0` instead of the patch to `v0.8.1`? I guess it doesn't make a big difference in the grand scheme of things.

liamsi:
Yeah, could have been. It's not breaking because the change is purely additive but 🤷🏻‍♂️

PR #58, opened 2022-05-29
Verify inclusion proofs with multiple leaves

evan-forbes:
building on #57, we also need to be able to verify inclusion proofs for arbitrary ranges of leaves. 

Note: this PR is API breaking and will require another version bump

### DIFF
--- a/fuzz_test.go
+++ b/fuzz_test.go
@@ -8,7 +8,7 @@ import (
 
 	"github.com/celestiaorg/nmt"
 	"github.com/celestiaorg/nmt/namespace"
-	"github.com/google/gofuzz"
+	fuzz "github.com/google/gofuzz"
 )
 
 func TestFuzzProveVerifyNameSpace(t *testing.T) {
@@ -72,7 +72,7 @@ func TestFuzzProveVerifyNameSpace(t *testing.T) {
 				if err != nil {
 					t.Fatalf("error on Prove(%v): %v", i, err)
 				}
-				if ok := singleItemProof.VerifyInclusion(hash, data[i][:size], data[i][size:], treeRoot); !ok {
+				if ok := singleItemProof.VerifyInclusion(hash, data[i][:size], [][]byte{data[i][size:]}, treeRoot); !ok {
 					t.Fatalf("expected VerifyInclusion() == true; data = %#v; proof = %#v", data[i], singleItemProof)
 				}
 				leafIdx++
--- a/proof.go
+++ b/proof.go
@@ -210,10 +210,7 @@ func (proof Proof) VerifyInclusion(h hash.Hash, nid namespace.ID, data [][]byte,
 	nth := NewNmtHasher(h, nid.Size(), proof.isMaxNamespaceIDIgnored)
 	hashes := make([][]byte, len(data))
 	for i, d := range data {
-		leafData := append(append(
-			make([]byte, 0, len(d)+len(nid)),
-			nid...),
-			d...)
+		leafData := append(append(make([]byte, 0, len(d)+len(nid)), nid...), d...)
 		hashes[i] = nth.HashLeaf(leafData)
 	}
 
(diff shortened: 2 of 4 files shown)


proof.go
```diff
@@ -206,10 +206,18 @@ func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID na
 	return bytes.Equal(tree.Root(), root)
 }
 
-func (proof Proof) VerifyInclusion(h hash.Hash, nid namespace.ID, data []byte, root []byte) bool {
+func (proof Proof) VerifyInclusion(h hash.Hash, nid namespace.ID, data [][]byte, root []byte) bool {
```
evan-forbes, line 209: This api is arbitrarily restricting proofs to the same namespace. While that fits our use case for the rest of the repos, the api could be simplified to match `tree.Push()` where the namespace is included in the data. This would simplify some other things as well, but I felt that was outside the scope of this PR.

adlerjohn:
Can you note in the PR description that this is a breaking API change?

proof.go
adlerjohn: Can you document this public function? I don't actually know what `[][]byte` is just by looking at it.

proof.go
```diff
@@ -206,10 +206,18 @@ func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID na
...
 	nth := NewNmtHasher(h, nid.Size(), proof.isMaxNamespaceIDIgnored)
-	leafData := append(nid, data...)
-	return proof.verifyLeafHashes(nth, false, nid, [][]byte{nth.HashLeaf(leafData)}, root)
+	hashes := make([][]byte, len(data))
+	for i, d := range data {
+		leafData := append(append(
+			make([]byte, 0, len(d)+len(nid)),
+			nid...),
+			d...)
```
adlerjohn, line 216: Formatting is weird.

proof.go
evan-forbes: sure thing https://github.com/celestiaorg/nmt/pull/58/commits/d40c9b5eb2eaa19f5307663d9b40b1622221248c

proof.go
```diff
@@ -206,10 +206,19 @@ func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID na
 	return bytes.Equal(tree.Root(), root)
 }
 
-func (proof Proof) VerifyInclusion(h hash.Hash, nid namespace.ID, data []byte, root []byte) bool {
+// VerifyInclusion checks that the inclusion proof is valid by using leaf data
+// and the provided proof to regenerate and compare the root. Note that the leaf
+// data should not contain the prefixed namespace, unlike the tree.Push method,
+// which takes prefixed data
```
adlerjohn, line 212:
```suggestion
// which takes prefixed data. All shares implicitly have the same namespace ID: `nid`.
```

proof.go
```diff
@@ -206,10 +206,19 @@ func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID na
 	return bytes.Equal(tree.Root(), root)
 }
 
-func (proof Proof) VerifyInclusion(h hash.Hash, nid namespace.ID, data []byte, root []byte) bool {
+// VerifyInclusion checks that the inclusion proof is valid by using leaf data
+// and the provided proof to regenerate and compare the root. Note that the leaf
+// data should not contain the prefixed namespace, unlike the tree.Push method,
+// which takes prefixed data. All leaves implicitly have the same namespace ID: `nid`.
+func (proof Proof) VerifyInclusion(h hash.Hash, nid namespace.ID, leaves [][]byte, root []byte) bool {
```
liamsi, line 213: An additional convenience method `VerifyNamespaceRange` or sth alike does not make sense?

liamsi (approved):

proof.go
evan-forbes: I can see that working as well, tbh I just didn't think about that and was just following the underlying library's API.

I want to merge to wrap this up, but I also don't want to lose this comment, so #59

fuzz_test.go
```diff
@@ -8,7 +8,7 @@ import (
 
 	"github.com/celestiaorg/nmt"
 	"github.com/celestiaorg/nmt/namespace"
-	"github.com/google/gofuzz"
+	fuzz "github.com/google/gofuzz"
```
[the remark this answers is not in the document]
evan-forbes: #60

Issue #63, opened 2022-07-25
Define license policy


rootulp:
Documentation for this package isn't displayed on pkg.go.dev because there is no license policy defined. See https://pkg.go.dev/license-policy

## Screenshot

<img width="1494" alt="Screen Shot 2022-07-25 at 5 17 12 PM" src="https://user-images.githubusercontent.com/3699047/180875626-4600b277-5d0b-4467-8a05-ea890e3206f7.png">

rootulp:
Apache License 2.0 
- https://github.com/celestiaorg/celestia-node/blob/main/LICENSE
- https://github.com/celestiaorg/celestia-app/blob/master/LICENSE
- https://github.com/celestiaorg/celestia-core/blob/v0.35.x-celestia/LICENSE
- https://github.com/celestiaorg/rsmt2d/blob/master/LICENSE
- https://github.com/celestiaorg/optimint/blob/main/LICENSE

MIT
- https://github.com/celestiaorg/smt/blob/master/LICENSE

Given Apache License 2.0 seems like the dominant license file for repos in celestiaorg, will opt to add that

PR #64, opened 2022-07-25
Add Apache License 2.0

rootulp:
Resolves https://github.com/celestiaorg/nmt/issues/63

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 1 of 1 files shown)


evan-forbes (commented):
seems fine to me, but I want to triple check with folks before adding. If so, should this be the default license added to every repo?

cc @adlerjohn @musalbas

adlerjohn (approved):

adlerjohn:
> should this be the default license added to every repo?

There's no "default" license, but every public repo should have an open-source license.

rootulp:
I don't have merge permissions

Screen Shot 2022-07-27 at 11 44 19 AM

but maybe we should wait for explicit approval from @musalbas pre merge

PR #65, opened 2022-09-05
Update example in README

rahulghangas:
Fix outdated example in README


adlerjohn (approved):
utACK, I'll trust you tested it

PR #67, opened 2022-09-05
Remove dependency on nebulous in verification

rahulghangas:
Use the same recursive approach as in `nmt.Root()` to verify leaf hashes
Removes the dependency on `Celestiaorg/merkletree` in proof verification

- [x] Implementation
- [x] Passing full testing

Fixes half of the issue mentioned in #15

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -167,11 +167,6 @@ func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID na
 		}
 	}
 
-	i := 1
-	for i < proof.end {
-		i *= 2
-	}
-
 	var computeRoot func(start, end int) []byte
 	computeRoot = func(start, end int) []byte {
 		switch end - start {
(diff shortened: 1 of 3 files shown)


proof.go
```diff
@@ -200,10 +167,47 @@ func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID na
 		}
 	}
 
-	// add remaining proof hashes after the last range ends
-	consumeUntil(math.MaxUint64)
+	var computeRoot func(start, end int) []byte
+	computeRoot = func(start, end int) []byte {
```
evan-forbes, line 171: [optional]

if I'm not mistaken, I think we could use the [one tendermint uses](https://github.com/tendermint/tendermint/blob/5013bc3f4a6d64dcc2bf02ccc002ebc9881c62e4/crypto/merkle/tree.go#L7-L21). If that's the case, I'd be okay with copy/pasting that code here and using it. This is an optional idea tho, this code appears to work fine.

evan-forbes (commented):
LGTM, I'd approve after resolving the comments from Ismail above

adlerjohn (commented):
same as above

PR #68, opened 2022-09-14
Remove dependency on nebulous in proof generation

rahulghangas:
Use the same recursive approach as in nmt.Root() to generate proof
Removes the dependency on Celestiaorg/merkletree in proof generation

- [x] Passing full testing
- [x] Fixes https://github.com/celestiaorg/nmt/issues/15


evan-forbes (commented):
LGTM, 

[blocking question] should we delete the entire internal package now that we're not using it?

liamsi (changes requested):
Agree with @evan-forbes we should delete the internal package entirely.

proof_test.go
```diff
@@ -3,14 +3,58 @@ package nmt
 import (
 	"bytes"
 	"crypto/sha256"
+	"io"
 	"testing"
 
 	"github.com/celestiaorg/merkletree"
```
liamsi, line 9: Ideally, we would remove this dependency entirely. But orthogonal to this PR.

rahulghangas: Done

rahulghangas:
Whoops, keep on forgetting you can only request one review. Sorry

proof_test.go
rahulghangas: Edit -- This cannot be removed, merkletree relies on io errors internally, which is wild!

rahulghangas: Anyway, this is only in a test, I think should be fine?

evan-forbes (approved):
LGTM

adlerjohn:
Poking @liamsi for re-review since changes requested.

nmt.go
```diff
@@ -127,12 +126,11 @@ func (n NamespacedMerkleTree) Prove(index int) (Proof, error) {
 func (n NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
 	isMaxNsIgnored := n.treeHasher.IsMaxNamespaceIDIgnored()
 	n.computeLeafHashesIfNecessary()
-	subTreeHasher := internal.NewCachedSubtreeHasher(n.leafHashes, n.treeHasher)
 	// TODO: store nodes and re-use the hashes instead recomputing parts of the tree here
-	proof, err := merkletree.BuildRangeProof(start, end, subTreeHasher)
-	if err != nil {
-		return NewEmptyRangeProof(isMaxNsIgnored), err
+	if start < 0 || start > end || start == end || end > len(n.leafHashes) {
```
adlerjohn, line 130: Elsewhere in the code you use `>=`, so for consistency

```suggestion
	if start < 0 || start >= end || end > len(n.leafHashes) {
```

nmt.go
```diff
@@ -171,27 +169,66 @@ func (n NamespacedMerkleTree) ProveNamespace(nID namespace.ID) (Proof, error) {
...
+func (n NamespacedMerkleTree) buildRangeProof(proofStart, proofEnd int) [][]byte {
+	proof := [][]byte{}
+	var recurse func(start, end int, includeNode bool) []byte
+	recurse = func(start, end int, includeNode bool) []byte {
+		if start >= len(n.leafHashes) {
+			return nil
+		}
+
+		// reached a leaf
+		if end-start == 1 {
```
adlerjohn, line 189: Did you run gofmt on this file? Doesn't look like we have it in CI, but this line looks sus.

rahulghangas: Actually, gofmt changes it to this form

rahulghangas:
bulld is failing due to codecov upload failure

adlerjohn (approved):
utACK

adlerjohn:
Poking @liamsi for re-review since changes requested.

rahulghangas:
@liamsi bump

rahulghangas:
@adlerjohn can be merged now?

adlerjohn (approved):

PR #76, opened 2022-10-17
prune: speed up using 1<<N instead of math.Pow(2, N)

### LINKED ISSUE #72: nmt: unnecessary and expensive powerOf2 exponentiation using math.Pow then uint(float64) cast yet could simply use left bit shifts aka 1<<N
This code here https://github.com/celestiaorg/nmt/blob/230d27f2d5d044a545b21a5a8f09abd73f7cc410/subrootpaths.go#L74
image
performs exponentiation for powers of 2 in a very expensive and unnecessary way, this simply could be 1<<N which makes for fast code and this is an articulation for example:
* uint(math.Pow(float64(2), float64(8))) == 1<<8
* uint(math.Pow(float64(2), float64(1))) == 1<<1
* uint(math.Pow(float64(2), float64(12))) == 1<<12

## Benchmarks
If you run these benchmarks with `go test -run=^$ -bench=Pow2 -count=10` then compare
```go
var sink any

func BenchmarkPow2Naive(b *testing.B) {
        benchmarkPow2(b, func(i int) uint {
                return uint(math.Pow(float64(2), float64(i)))
        })
}

func BenchmarkPow2BitShift(b *testing.B) {
        benchmarkPow2(b, func(i int) uint {
                return 1 << i
        })
}

var cases = []struct {
        i    int
        want uint
}{
        {0, 1},
        {1, 2},
        {2, 4},
        {3, 8},
        {4, 16},
        {5, 32},
        {12, 4096},
        {20, 1048576},
        {24, 16777216},
        {31, 2147483648},
        {63, 9223372036854775808},
}

func benchmarkPow2(b *testing.B, fn func(int) uint) {
        b.ReportAllocs()
        for i := 0; i < b.N; i++ {
                for _, tc := range cases {
                        n := fn(tc.i)
                        if n != tc.want {
                                b.Fatalf("Bad result: got=%d want=%d", n, tc.want)
                        }
                        sink = n
                }
        }

        if sink == nil {
                b.Fatal("Benchmark did not run")
        }
        sink = (any)(nil)
}
```

## Results
```shell
$ benchstat before.txt after.txt 
name    old time/op    new time/op    delta
Pow2-8     322ns ± 2%     114ns ± 1%  -64.69%  (p=0.000 n=9+8)

name    old alloc/op   new alloc/op   delta
Pow2-8     40.0B ± 0%     40.0B ± 0%     ~     (all equal)

name    old allocs/op  new allocs/op  delta
Pow2-8      5.00 ± 0%      5.00 ± 0%     ~     (all equal)
```

Kind cc @elias-orijtech @kirbyquerby @willpoint @jhusdero @musalbas @liamsi

odeke-em:
This change fixes the use of expensive code that used math.Pow(float(2), N) but really that's expensive code that could be simplified by using left bitshifts.

The improvements in speed show in benchmarks:

```shell
$ benchstat before.txt after.txt
name    old time/op    new time/op    delta
Pow2-8     322ns ± 2%     114ns ± 1%  -64.69%  (p=0.000 n=9+8)

name    old alloc/op   new alloc/op   delta
Pow2-8     40.0B ± 0%     40.0B ± 0%     ~     (all equal)

name    old allocs/op  new allocs/op  delta
Pow2-8      5.00 ± 0%      5.00 ± 0%     ~     (all equal)
```

Fixes #72

### DIFF
--- a/subrootpaths.go
+++ b/subrootpaths.go
@@ -71,7 +71,9 @@ func prune(idxStart uint, idxEnd uint, maxWidth uint) [][]int {
 	rightTraversed := false
 
 	for i := treeDepth - 1; i >= 0 && capturedSpan < idxEnd; i-- {
-		nodeSpan := uint(math.Pow(float64(2), float64(treeDepth-i)))
+		// nodeSpan is 2**(treeDepth-i) == 1<<(treeDepth-i)
+		// Please see: https://github.com/celestiaorg/nmt/issues/72
+		nodeSpan := uint(1 << (treeDepth - i))
 		if pathStart[i] == 0 {
 			// if nodespan is less than end index, continue traversing upwards
 			lastNode := nodeSpan + idxStart - 1


liamsi (approved):
Thanks!

PR #79, opened 2022-10-17
nmt: sanity check addition results to avoid overflows

odeke-em:
This change cautiously checks addition results to prevent integer overflows. Also checks errors from left and right GetSubrootPaths generation to avoid any future runtime panics. While here, also slightly tweaked table tests for TestPathGeneration to use subtests and report which case failed exactly.

### DIFF
--- a/subrootpaths.go
+++ b/subrootpaths.go
@@ -9,6 +9,8 @@ var (
 	srpNotPowerOf2       = errors.New("GetSubrootPaths: Supplied square size is not a power of 2")
 	srpInvalidShareCount = errors.New("GetSubrootPaths: Can't compute path for 0 share count slice")
 	srpPastSquareSize    = errors.New("GetSubrootPaths: Share slice can't be past the square size")
+
+	srpInvalidIdxStartShareCountSum = errors.New("invalid idxStart+shareCount value")
 )
 
 // merkle path to a node is equivalent to the index's binary representation
@@ -123,10 +125,6 @@ func prune(idxStart uint, idxEnd uint, maxWidth uint) [][]int {
 // An empty entry in the top level list means the shares span that entire row and so
 // the root for that segment of shares is equivalent to the row root.
 func GetSubrootPaths(squareSize uint, idxStart uint, shareCount uint) ([][][]int, error) {
-
-	var paths [][]int
-	var top [][][]int
-
 	// check squareSize is at least 2 and that it's
 	// a power of 2 by checking that only 1 bit is on
 	if squareSize < 2 || bits.OnesCount(squareSize) != 1 {
@@ -138,10 +136,15 @@ func GetSubrootPaths(squareSize uint, idxStart uint, shareCount uint) ([][][]int
 		return nil, srpInvalidShareCount
 	}
 
+	isn := idxStart + shareCount
+	if isn < idxStart || isn < shareCount {
+		return nil, srpInvalidIdxStartShareCountSum
+	}
+
 	shares := squareSize * squareSize
 
 	// sanity checking
-	if idxStart >= shares || idxStart+shareCount > shares {
+	if idxStart >= shares || isn > shares {
 		return nil, srpPastSquareSize
 	}
 
@@ -155,6 +158,9 @@ func GetSubrootPaths(squareSize uint, idxStart uint, shareCount uint) ([][][]int
 	shareStart := idxStart % squareSize
 	shareEnd := (idxStart + shareCount - 1) % squareSize
 
+	var paths [][]int
+	var top [][][]int
+
 	// if the count is one, just return the subdivided start path
 	if shareCount == 1 {
 		return append(top, append(paths, subdivide(shareStart, squareSize))), nil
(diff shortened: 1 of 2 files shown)


odeke-em:
Kind /cc @liamsi @musalbas @adlerjohn

subrootpaths.go
```diff
@@ -133,8 +127,15 @@ func GetSubrootPaths(squareSize uint, idxStart uint, shareCount uint) ([][][]int
 		return nil, srpInvalidShareCount
 	}
 
+	isn := idxStart + shareCount
+	if isn < idxStart || isn < shareCount {
+		return nil, errors.New("invalid idxStart+shareCount value")
```
liamsi, line 132: The max share count possible in Celestia is bound by the maxchunks of the erasure coding library. So this check, while cool, is not a code path that will ever be hit in practice. If you look at nmt independently as a library it makes more sense though.

liamsi (approved):

subrootpaths.go
odeke-em: Much safer to not bake those assumptions into a library that'll be open to everyone to use. I was able to hit that point from trying to attack this library. However I do understand the context.

subrootpaths.go
```diff
@@ -133,8 +127,15 @@ func GetSubrootPaths(squareSize uint, idxStart uint, shareCount uint) ([][][]int
 		return nil, srpInvalidShareCount
 	}
 
+	isn := idxStart + shareCount
```
rootulp, line 130: [question] what does isn stand for? Since it isn't immediately obvious, proposal to rename to something more verbose

subrootpaths.go
rootulp, line 132: [optional] it may help future developers if this error string included additional debugging information

```suggestion
		return nil, fmt.Sprintf("invalid isn %v must be less than idxStart %v and shareCount %v", isn, idxStart, shareCount)
```

If this proposal isn't accepted, the existing static error string may be hoisted to line 13 of this file so that all error strings are defined in one location for this file.

rootulp: [blocking] is it possible to include a unit test that demonstrates this error is returned under the expected circumstances?

subrootpaths_test.go
```diff
@@ -133,10 +133,13 @@ func TestPathGeneration(t *testing.T) {
 	}
 
 	for _, tc := range tests {
-		paths, err := GetSubrootPaths(tc.input.squareSize, tc.input.startNode, tc.input.length)
-		if !reflect.DeepEqual(pathResult(paths), tc.want) {
-			t.Fatalf(`GetSubrootPaths(%v) = %v, %v, want %v - rationale: %v`, tc.input, paths, err, tc.want, tc.desc)
-		}
+		tc := tc
+		t.Run(tc.desc, func(t *testing.T) {
```
rootulp, line 137: 👍 thanks for updating this to report the test case description on failures

subrootpaths.go
odeke-em: I am purposefully not sending in a PR given it can be exploited, I had spoken to y'all offline for responsible disclosure of the bug :-)

subrootpaths_test.go
odeke-em: Ack and thanks!

subrootpaths.go
odeke-em: It's just a variable combo of idxStart and shareCount: isn

I don't really know what contextual name to give it :-)

subrootpaths.go
odeke-em: I've hoisted the error out as you requested @rootulp

subrootpaths.go
```diff
@@ -9,6 +9,8 @@ var (
 	srpNotPowerOf2       = errors.New("GetSubrootPaths: Supplied square size is not a power of 2")
 	srpInvalidShareCount = errors.New("GetSubrootPaths: Can't compute path for 0 share count slice")
 	srpPastSquareSize    = errors.New("GetSubrootPaths: Share slice can't be past the square size")
+
+	srpInvalidIdxStartShareCountSum = errors.New("invalid idxStart+shareCount value")
```
rootulp, line 13: [optional] if the other rename suggestion is accepted

```suggestion
	srpInvalidIdxEnd = errors.New("GetSubrootPaths: idxEnd must be larger than idxStart and shareCount")
```

subrootpaths.go
rootulp: > I am purposefully not sending in a PR given it can be exploited

I don't think we should be concerned about adding a unit test for this error. A unit test gives maintainers greater confidence that the bug has been addressed.

If an attacker wants to exploit this integer overflow, they would be able to do so irrespective of this unit test.

rootulp (approved):
I left two optional renames and a rationale for why I think a unit test is still valubale here. If the comments aren't addressed, I'll create a follow-up PR that addresses them so approving to unblock. Thanks for identifying the issue and proposing this fix @odeke-em !

PR #80, opened 2022-10-20
Simplify GetSubrootPaths

elias-orijtech:
CC @odeke-em

### DIFF
--- a/subrootpaths.go
+++ b/subrootpaths.go
@@ -128,8 +128,6 @@ func GetSubrootPaths(squareSize uint, idxStart uint, shareCount uint) ([][][]int
 	var paths [][]int
 	var top [][][]int
 
-	shares := squareSize * squareSize
-
 	// check squareSize is at least 2 and that it's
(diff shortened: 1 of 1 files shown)


subrootpaths.go
```diff
@@ -141,22 +139,25 @@ func GetSubrootPaths(squareSize uint, idxStart uint, shareCount uint) ([][][]int
...
 	// sanity checking
 	if idxStart >= shares || idxStart+shareCount > shares {
 		return nil, srpPastSquareSize
 	}
 
-	// adjust for 0 index
-	shareCount = shareCount - 1
-
-	startRow := int(math.Floor(float64(idxStart) / float64(squareSize)))
-	closingRow := int(math.Ceil(float64(idxStart+shareCount) / float64(squareSize)))
```
liamsi, line 153: The linter noticed that the math package is still imported (unnecessarily) btw.

elias-orijtech: Done.

PR #82, opened 2022-10-31
test: add fuzzers to prevent vulnerability regressions

odeke-em:
Adds a test and fuzzers that will help catch vulnerabilities that could relapse.


cmwaters:
Closing this as being outdated unfortunately. We have added some fuzzing since this PR but would advocate for more where appropriate.

PR #90, opened 2022-12-16
Bug fix: reject forged proofs of absence. Fixes (#86)

### LINKED ISSUE #86: Bug: Proofs of Absence can be Forged

preston-evans98:
- Reject empty "proofs" of absence, except in the special case where the root doesn't cover the namespace being proved
- Add test case to catch regression
- Add test case to ensure that the absence of the zero namespace can be proved against the empty tree

closes #86

## Checklist

### DIFF
--- a/nmt_test.go
+++ b/nmt_test.go
@@ -642,8 +642,8 @@ func TestNMT_forgedNamespaceEmptinessProof(t *testing.T) {
 		isMaxNamespaceIDIgnored: true,
 	}
 
-	forged_proof_succes := forgedProof.VerifyNamespace(sha256.New(), namespace.ID{1}, [][]byte{}, root)
-	if forged_proof_succes {
+	forgedProofSuccess := forgedProof.VerifyNamespace(sha256.New(), namespace.ID{1}, [][]byte{}, root)
+	if forgedProofSuccess {
 		t.Fatalf("Successfully verified proof that non-empty namespace was empty")
 	}
 }
--- a/nmt_test.go
+++ b/nmt_test.go
@@ -596,6 +596,58 @@ func TestNamespacedMerkleTree_calculateAbsenceIndex_Panic(t *testing.T) {
 	}
 }
 
+// This test checks for a regression of https://github.com/celestiaorg/nmt/issues/86
+func TestNMT_absenceProofOfZeroNamespace_InEmptyTree(t *testing.T) {
+	tree := New(sha256.New(), NamespaceIDSize(1))
+	root := tree.Root()
+	emptyleaves, proof, err := tree.GetWithProof(namespace.ID{0})
+	if err != nil {
+		t.Fatalf("GetWithProof()  could not get namespace{0}. err: %v ", err)
+	}
+	if len(emptyleaves) != 0 {
+		t.Fatalf("Get(namespace.ID{0}) should have returned no leaves but returned %v", emptyleaves)
+	}
+	if !proof.VerifyNamespace(sha256.New(), namespace.ID{0}, emptyleaves, root) {
+		t.Fatalf("Could not verify proof of absence of namespace zero in empty tree")
+	}
+}
+
+// This test checks for a regression of https://github.com/celestiaorg/nmt/issues/86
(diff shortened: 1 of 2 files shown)


nmt_test.go
```diff
@@ -596,6 +596,58 @@ func TestNamespacedMerkleTree_calculateAbsenceIndex_Panic(t *testing.T) {
...
+	forgedProof := Proof{
+		start:                   0,
+		end:                     0,
+		nodes:                   [][]byte{},
+		leafHash:                []byte{},
+		isMaxNamespaceIDIgnored: true,
+	}
+
+	forged_proof_succes := forgedProof.VerifyNamespace(sha256.New(), namespace.ID{1}, [][]byte{}, root)
+	if forged_proof_succes {
```
liamsi, line 646: Just a minor nit (snake_case is very unidiomatic in golang and there is a typo: s/succes/success):
```suggestion
	forgedProofSuccess := forgedProof.VerifyNamespace(sha256.New(), namespace.ID{1}, [][]byte{}, root)
	if forgedProofSuccess {
```

liamsi (commented):
Tests and logic look great. Left one minor nit.

Orthogonal to this PR we should probably ensure that the empty-tree case is always handled correctly in all the other existing table-driven tests as well.

liamsi (approved):
🙏🏼 🚀 ❤️

Issue #87, opened 2022-12-15
Verification logic does not ensure lexicographical ordering


liamsi:
The NMT guarantees lexicographical ordering and hence generated proofs follow this ordering as well. The verification logic does not enforce this though currently. This might not be a big deal as verification would fail if ordering is different though it makes it harder to reason about this and it is obviously better to enforce ordering there too.

@musalbas suggested to enforce ordering in the hasher directly. While it might seem better to follow the current pattern and fail much earlier (when adding data to the tree and not only later when hashing/creating the root), having the ordering enforced while hashing is better from a consistency perspective. it would also immediately resolve the ordering guarantee during verification as well.

liamsi:
The only downside I can think of, when incorporating the ordering into the hasher, is that it will have to happen on every inner node (instead of currently every leaf only), which will make proving and verification slower (probs negligible but should be confirmed in a benchmark too).

liamsi:
Note that the hasher is also exposed directly and indpependently of an NMT here: https://github.com/celestiaorg/nmt/blob/7d31915d17e5019abef8bdea89a3182842331cba/hasher.go#L22-L54

We should remove this code (looks like celestia-node is the only user of it and even that is about to change [here ](https://github.com/celestiaorg/celestia-node/pull/1498/files). If we keep these functions there we actually _must_ enforce the ordering in the hasher as otherwise this part of the public API can be misused.

liamsi:
closing in favour of https://github.com/celestiaorg/nmt/issues/95 (more actionable)

PR #91, opened 2023-01-03
feat!: implementing `hash.Hash` interface

distractedm1nd:
## Overview

Resolves a TODO from celestia-node:
https://github.com/celestiaorg/celestia-node/blob/8f13efab8aa314d52708dea8873eea70027745cb/share/ipld/namespace_hasher.go#L21-L33

Motivation: Instead of only providing utilities that require knowing if a value is an inner node or a leaf to hash, implement the `hash.Hash` interface so that `Write`/`Sum` can be used generally over both node types.

This PR is breaking as it changes the `Hasher` constructor to include share size.

To be discussed:
- [ ] Is this even desired? We want it in node but hasn't been discussed elsewhere AFAIK
- [ ] This PR introduces testify a dep, I can rewrite the tests to exclude this if unwanted
- [ ] Why is NamespaceLen exported, and has a getter in addition? Should `shareSize` also be exported then?

## Checklist

- [X] New and updated code has appropriate documentation
- [X] New and updated code has new and/or updated testing


liamsi (commented):
Looks great! Will do another pass tomorrow morning. 
Linking this issue as it is somewhat related: https://github.com/celestiaorg/nmt/issues/87

nmt.go
```diff
@@ -11,6 +11,12 @@ import (
 	"github.com/celestiaorg/nmt/namespace"
 )
 
+const (
+	DefaultNamespaceIDLen = 8
+	DefaultShareSize      = 512
```
Wondertan, line 16: Might be worth calling it LeafSiae as Share is a concept outside of the NMT and there is a notion of Leaf already in NMT.

hasher.go
```diff
@@ -69,18 +74,76 @@ func (n *Hasher) NamespaceSize() namespace.IDSize {
...
+
+	ln := len(data)
+	switch ln {
+	default:
+		return 0, fmt.Errorf("invalid data size written to the hasher, len: %v", ln)
+	// inner nodes are made up of the nmt hashes of the left and right children
+	case n.Size() * 2:
+		n.tp = NodePrefix
+	// leaf nodes contain the namespace length and a share
+	case int(n.NamespaceLen) + n.shareSize:
```
Wondertan, line 110: We might consider making leaf share a default case, then we wouldn't need to pass shareSize. 

It's ok to unlimit the share/leaf size as NMT can be used outside of the celestia'a use case, where leafs could have different sizes.

There is no security issues with this as integrity is checked anyway. As I can see so far no other panics would be introduced, but we should triple check that if decide leaf to be a default case.

Wondertan:
Doesn't diverge from implementation in celestia-node so LGTM

liamsi (commented):
Agree with @Wondertan's comments (particularly using LeafSize instead of ShareSize but also not enforcing a fixed share size here). Otherwise LGTM 👏🏼

Wondertan (approved):
Thanks for moving this ❤️

Wondertan:
I lost my nmt approval perms😢

hasher.go
```diff
@@ -71,16 +74,70 @@ func (n *Hasher) NamespaceSize() namespace.IDSize {
...
+// Size returns the number of bytes Sum will return.
+func (n *Hasher) Size() int {
+	return n.baseHasher.Size() + int(n.NamespaceLen)*2
+}
+
+// Write writes the namespaced data to be hashed.
+//
+// Requires data of fixed size to match leaf or inner NMT nodes.
+// Only a single write is allowed.
+func (n *Hasher) Write(data []byte) (int, error) {
```
adlerjohn, line 93: Does this implement another interface? If so, can you document? If it doesn't implement an interface then I'd say just don't allow more than one call to `Write` by construction.

nmt.go
```diff
@@ -23,6 +28,7 @@ type NodeVisitorFn = func(hash []byte, children ...[]byte)
 type Options struct {
 	InitialCapacity    int
 	NamespaceIDSize    namespace.IDSize
+	ShareSize          int
```
Wondertan, line 31: Randomly found a relic that we can remove

hasher.go
Wondertan: It implements `hash.Hash`, which implements `io.Writer`.

adlerjohn (approved):

Issue #71, opened 2022-10-07
namespace: PrefixedData/PrefixedData8 are easy to misuse


elias-orijtech:
The `PrefixedData` and `PrefixedData8` are both byte slices:

```go
// PrefixedData simply represents a slice of bytes which consists of
// a namespace.ID and raw data.
// The user has to guarantee that the bytes are valid namespace prefixed data.
// Go's type system does not allow enforcing the structure we want:
// [namespaceID, rawData ...], especially as this type does not expect any
// particular size for the namespace.
type PrefixedData []byte

// PrefixedData8 like PrefixedData is just a slice of bytes.
// It assumes that the slice it represents is at least 8 bytes.
// This assumption is not enforced by the type system though.
type PrefixedData8 []byte
```

The definitions are problematic, because (1) casting byte slices may violate the type invariants (as pointed out in the comments) (2) they're mutable (3) PrefixedData has no way to indicate the namespace identifier length. For an example where mutability is a problem, see

https://github.com/celestiaorg/nmt/blob/62742433f65a37daf889ee563b0ee117c83e9664/nmt.go#L253-L257

where it is not clear to the caller that the parameter `namespacedData` is retained by the `Push` method.

The missing namespace length is problematic for `validateAndExtractNamespace`,

https://github.com/celestiaorg/nmt/blob/62742433f65a37daf889ee563b0ee117c83e9664/nmt.go#L336-L340

because it can only validate that the namespace+data total length is at least as long as the expected namespace length.

Perhaps the definitions of `PrefixedData` and `PrefixedData8` are caused by wanting to support zero-copy and zero-allocation use, where the user generates `PrefixedData` values by slicing existing byte slices. If so, the behaviour of `Push` should document that argument values are retained.

Otherwise, I recommend changing the definition to hide the internal structure, and add validating constructors. There are several possibilites:

```go
type PrefixedData struct {
    // namespace and data can be sliced from the same
    // underlying []byte allocation.
    namespace []byte
    data []byte
}

func NewPrefixedData(namespace, data []byte) PrefixedData {
    bytes := make([]byte, len(namespace)+len(data))
    d := PrefixedData{
        namespace: bytes[:len(namespace)],
        data: bytes[len(namespace)],
    }
    copy(d.namespace, namespace)
    copy(d.data, data)
    return d
}
```

or equivalently

```go
type PrefixedData struct {
    namespaceAndData []byte
    namespaceLen int
}
```

@odeke-em

[...]

PR #94, opened 2023-02-01
NMT Review

staheri14:
This is not a formal pull request for review, but rather a place to share my findings, questions, and suggestions regarding the implementation of the NMT as part of the following EPIC (https://github.com/celestiaorg/celestia-app/issues/1296). I am conducting a thorough review of the NMT implementation and design. 
The current content of the PR will undertake more updates as I am continuously writing and updating the code documentation, and adding more questions and suggestion. I am using this PR only as a place to communicate my thoughts. 

To enhance the code documentation, I have added descriptions for some of the functions that were previously lacking proper documentation. I would appreciate if you could take a look and provide feedback, @liamsi. 

My suggestions and questions are annotated as `TODO [Me]` to distinguish them from existing annotations in the code. I will continue to add comments and questions as I proceed with my review. Upon resolution of any questions, I will replace the annotations with proper documentation. If my suggestions are approved, I will open GitHub issues to implement them in subsequent pull requests.

The primary objectives of my code review are to:

1. Understand the proof generation algorithm and the structure of the `Proof`, particularly the order of nodes retrieved in the "nodes" field of the `Proof` structure and how to interpret the other fields in this struct.
2. Evaluate the verification procedure.
3. Identify any opportunities for security or performance improvement.
4. Improve the code documentation and readability, which would also be beneficial for future maintainers.

cc: @evan-forbes

### DIFF
--- a/namespace/data.go
+++ b/namespace/data.go
@@ -6,6 +6,7 @@ package namespace
 // Go's type system does not allow enforcing the structure we want:
 // [namespaceID, rawData ...], especially as this type does not expect any
 // particular size for the namespace.
+// TODO [Me] Shouldn't we specify that the first 8 bytes represent the namespace.ID
 type PrefixedData []byte
 
 // PrefixedData8 like PrefixedData is just a slice of bytes.
--- a/nmt.go
+++ b/nmt.go
@@ -337,7 +337,6 @@ func (n *NamespacedMerkleTree) calculateAbsenceIndex(nID namespace.ID) int {
 // foundInRange returns true, together with the starting and ending indices of the leaves in the name space tree whose namespace ID matches the given nID.
 // if no leaves is found, foundInRange returns false.
 // Note that the ending index is non-inclusive
-// TODO [Me] we could also incorporate the logic to handle out of range
 func (n *NamespacedMerkleTree) foundInRange(nID namespace.ID) (bool, int, int) {
 	// This is a faster version of this code snippet:
 	// https://github.com/celestiaorg/celestiaorg-prototype/blob/2aeca6f55ad389b9d68034a0a7038f80a8d2982e/simpleblock.go#L106-L117
--- a/nmt.go
+++ b/nmt.go
@@ -208,9 +208,9 @@ func (n *NamespacedMerkleTree) ProveNamespace(nID namespace.ID) (Proof, error) {
 	return NewAbsenceProof(proofStart, proofEnd, proof, n.leafHashes[proofStart], isMaxNsIgnored), nil
 }
 
-// buildRangeProof returns the nodes (as byte slices) in the audit path(inclusion/range proof/ siblings) of the supplied range i.e.,
-// [proofStart, proofEnd) where proofEnd is non-inclusive
-// The nodes are ordered according to in order traversal of the namespaced tree (TODO [Me] I need to double-check the order)
+// buildRangeProof returns the nodes (as byte slices) in the range proof of the supplied range i.e.,
+// [proofStart, proofEnd) where proofEnd is non-inclusive.
+// The nodes are ordered according to in order traversal of the namespaced tree. (TODO [Me] I need to double-check the order)
 func (n *NamespacedMerkleTree) buildRangeProof(proofStart, proofEnd int) [][]byte {
 	// TODO a more secure way would be to make it slice of fixed size arrays i.e., the hash output size
 	proof := [][]byte{} // it is the list of nodes hashes (as byte slices) with no index
--- a/nmt.go
+++ b/nmt.go
@@ -208,7 +208,6 @@ func (n *NamespacedMerkleTree) ProveNamespace(nID namespace.ID) (Proof, error) {
 	return NewAbsenceProof(proofStart, proofEnd, proof, n.leafHashes[proofStart], isMaxNsIgnored), nil
 }
 
-// TODO [Me] add a function description about how to parse the output [][]byte
 // buildRangeProof returns the nodes (as byte slices) in the audit path(inclusion/range proof/ siblings) of the supplied range i.e.,
 // [proofStart, proofEnd) where proofEnd is non-inclusive
 // The nodes are ordered according to in order traversal of the namespaced tree (TODO [Me] I need to double-check the order)
--- a/nmt.go
+++ b/nmt.go
@@ -149,13 +149,16 @@ func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
 
 // ProveNamespace returns a range proof for the given NamespaceID.
 //
-// case 1) In the case (nID < n.minNID) or (n.maxNID < nID) we do not
-// generate any range proof, instead we return an empty range (0,0) to
-// indicate that this namespace is not contained in the tree.
+// case 1) If the namespace nID is out of the range of the tree's min and max namespace
+// i.e., (nID < n.minNID) or (n.maxNID < nID)
+//  we do not generate any range proof, instead we return an empty Proof with the range (0,0) i.e.,
+// Proof.start = 0 and Proof.end = 0
+// to indicate that this namespace is not contained in the tree.
 //
-// case 2) If the tree does not have any entries with the given Namespace ID,
-// but the namespace is within the range of the tree's min and max namespace,
-// this will be proven by returning the (namespaced or rather flagged)
+// case 2) If the namespace nID is within the range of the tree's min and max namespace
+// i.e., n.minNID<= n.ID <=n.maxNID
+// and the tree does not have any entries with the given Namespace ID nID,
+// this will be proven by returning the inclusion/range Proof of the (namespaced or rather flagged)
 // hash of the leaf that is in the range instead of the namespace.
 //
 // case 3) In case the underlying tree contains leaves with the given namespace
@@ -299,6 +302,9 @@ func (n *NamespacedMerkleTree) GetWithProof(nID namespace.ID) ([][]byte, Proof, 
 	return data, proof, err
 }
 
+// return the index of a leaf with the largest namespace ID which is smaller than nID
+// in case there are multiple leaves with this property, then the index of the highest one
+// is returned
 func (n *NamespacedMerkleTree) calculateAbsenceIndex(nID namespace.ID) int {
 	nidSize := n.treeHasher.NamespaceSize()
 	var prevLeaf []byte
--- a/nmt.go
+++ b/nmt.go
@@ -141,7 +141,8 @@ func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
 	isMaxNsIgnored := n.treeHasher.IsMaxNamespaceIDIgnored()
 	n.computeLeafHashesIfNecessary()
 	// TODO: store nodes and re-use the hashes instead recomputing parts of the tree here
-	if start < 0 || start >= end || end > len(n.leafHashes) {
+	if start < 0 || start >= end || end > len(n.
+		leafHashes) {
 		return NewEmptyRangeProof(isMaxNsIgnored), ErrInvalidRange
 	}
 	proof := n.buildRangeProof(start, end)
@@ -176,6 +177,9 @@ func (n *NamespacedMerkleTree) ProveNamespace(nID namespace.ID) (Proof, error) {
 	if nID.Less(n.minNID) || n.maxNID.Less(nID) { // TODO [Me] we could move this entire if block inside the `foundInRange` function
 		// TODO [Me] Shouldn't we instead return the first or the last node in the tree as the exclusion proof?
 		// TODO [Me] although I think this current logic is based on the premise that the root of the tree is trusted
+		// TODO [Me] this check is redundant,
+		//  it is already being checked in ProveRange,
+		//  we may want to unite the two
 		return NewEmptyRangeProof(isMaxNsIgnored), nil
 	}
 
--- a/nmt.go
+++ b/nmt.go
@@ -404,7 +404,7 @@ func (n *NamespacedMerkleTree) computeRoot(start, end int) []byte {
 	}
 }
 
-// getSplitPoint returns the largest power of 2 less than length
+// getSplitPoint returns the largest power of 2 less than the length
 func getSplitPoint(length int) int {
 	if length < 1 {
 		panic("Trying to split a tree with size < 1")
--- a/nmt.go
+++ b/nmt.go
@@ -248,9 +248,9 @@ func (n *NamespacedMerkleTree) buildRangeProof(proofStart, proofEnd int) [][]byt
 			newIncludeNode = false
 		}
 
-		// TODO [Me] Optimization idea: if includeNode is true and [start, end) are within the [proofStart, proofEnd), then we do not need to
+		// TODO [Me] Optimization idea: if includeNode is true and [start, end) are within the [proofStart, proofEnd) range, then we do not need to
 		// TODO [Me] split further, as neither the current subtree nor any of its left and right subtrees are needed for the proof
-		
+
 		// recursively get left and right subtree
 		k := getSplitPoint(end - start)
 
@@ -355,6 +355,7 @@ func (n *NamespacedMerkleTree) Push(namespacedData namespace.PrefixedData) error
 
 	// update relevant "caches":
 	n.leaves = append(n.leaves, namespacedData)
+	// TODO [Me] why not calculating the hash right away and add it to n.leafHashes
 	n.updateNamespaceRanges()
 	n.updateMinMaxID(nID)
 	n.rawRoot = nil
--- a/nmt.go
+++ b/nmt.go
@@ -154,7 +154,9 @@ func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
 //
 // case 1) If the namespace nID is out of the range of the tree's min and max namespace
 // i.e., (nID < n.minNID) or (n.maxNID < nID)
-//  we do not generate any range proof, instead we return an empty Proof with the range (0,0) i.e.,
+//
+//	we do not generate any range proof, instead we return an empty Proof with the range (0,0) i.e.,
+//
 // Proof.start = 0 and Proof.end = 0
 // to indicate that this namespace is not contained in the tree.
 //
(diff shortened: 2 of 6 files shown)


namespace/data.go
```diff
@@ -6,6 +6,7 @@ package namespace
 // Go's type system does not allow enforcing the structure we want:
 // [namespaceID, rawData ...], especially as this type does not expect any
 // particular size for the namespace.
+// TODO [Me] Shouldn't we specify that the first 8 bytes represent the namespace.ID
 type PrefixedData []byte
```
liamsi, line 10: Yeah, the issue with this is that the library wants to support varying namespace sizes (not only 8 bytes necessarily). 

We could instead:
 - just not use a type alias for byte and simply accept that this is even easier to misuse (would have the nice side effect that the user would not need to cast between bytes and PrefixedData at all)
 - use a dedicated type; e.g. one of the two options here: https://github.com/celestiaorg/nmt/issues/71

nmt.go
```diff
@@ -96,6 +99,7 @@ type NamespacedMerkleTree struct {
 // and for the given namespace size (number of bytes).
 // If the namespace size is 0 this corresponds to a regular non-namespaced
 // Merkle tree.
+// TODO [Me] Why not using a fixed namespace size? i.e., the DefaultNamespaceIDLen, is the variable size for the space efficiency?
```
liamsi, line 102: We might move to a larger size in the future. Also, the idea was that this library could also be used independent from Celestia's specific use-case and that an NMT could be useful by itself. That said, I am not of aware of any use case outside of celestia currently.

proof.go
```diff
@@ -113,25 +116,27 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, data [][]byte,
...
 			}
-			gotLeafNid := namespace.ID(gotLeaf[:nIDLen])
+			gotLeafNid := namespace.ID(gotLeaf[:nIDLen]) // TODO [Me] a helper function
 			if !gotLeafNid.Equal(nID) {
-				// conflicting namespace IDs in data
+				// conflicting namespace IDs in leaves
 				return false
 			}
-			leafData := append(gotLeafNid, gotLeaf[nIDLen:]...)
+			leafData := append(gotLeafNid, gotLeaf[nIDLen:]...) // TODO why not just passing the leaf? isn't it the same?
```
liamsi, line 139: I think you are right!

nmt.go
```diff
@@ -142,71 +149,119 @@ func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
...
 		proofEnd = proofStart + 1
 	}
-	// At this point we either found the namespace in the tree or calculated
+
+	// case 3)
+	// At this point we either found leaves with the namespace nID in the tree or calculated
 	// the range it would be in (to generate a proof of absence and to return
 	// the corresponding leaf hashes).
-	n.computeLeafHashesIfNecessary()
+	n.computeLeafHashesIfNecessary()                 // TODO [Me] Why it is needed? cannot we make sure the leaves hashes are calculated as soon as a data is pushed to the tree?
```
liamsi, line 196: we could

nmt.go
```diff
@@ -304,6 +365,7 @@ func (n *NamespacedMerkleTree) Push(namespacedData namespace.PrefixedData) error
 
 	// update relevant "caches":
 	n.leaves = append(n.leaves, namespacedData)
+	// TODO [Me] why not calculating the hash right away and add it to n.leafHashes
```
liamsi, line 368: I do not remember if there was any reason for not doing that. Maybe the benchmarks where favourable for hashing all at the end? Though I do not see how that could be more efficient at all.

proof.go
```diff
@@ -27,7 +27,7 @@ type Proof struct {
 	leafHash []byte
 	// isMaxNamespaceIDIgnored is set to true if the tree from which
 	// this Proof was generated from is initialized with
-	// Options.IgnoreMaxNamespace == true.
+	// Options.IgnoreMaxNamespace == true. // TODO [Me]? not sure about the usage of this
```
liamsi, line 30: replied in the notion doc about this

proof.go
```diff
@@ -113,25 +116,27 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, data [][]byte,
...
 	if proof.IsOfAbsence() {
 		gotLeafHashes = append(gotLeafHashes, proof.leafHash)
 	} else {
-		// collect leaf hashes from provided data and
+		// collect leaf hashes from provided leaves and
 		// do some sanity checks:
 		hashLeafFunc := nth.HashLeaf
-		for _, gotLeaf := range data {
+		for _, gotLeaf := range leaves {
+			// TODO [Me] can be converted to something like isNameSpaceIDPrefixed()
```
liamsi, line 129: Would increase readability. I'm in favour of this idea.

nmt.go
```diff
@@ -115,6 +119,7 @@ func New(h hash.Hash, setters ...Option) *NamespacedMerkleTree {
 		leaves:          make([][]byte, 0, opts.InitialCapacity),
 		leafHashes:      make([][]byte, 0, opts.InitialCapacity),
 		namespaceRanges: make(map[string]leafRange),
+		// TODO [Me] Shouldn't minNID be populated by `0x00`?
```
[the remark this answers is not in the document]
staheri14: I see, then, does it mean that the `minID` is actually equal to `0xFF` which corresponds to the Max namespace ID?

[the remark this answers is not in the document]
staheri14: > That said, I think we can get rid of these two fields entirely!

Noted!

proof.go
```diff
@@ -94,17 +94,20 @@ func NewAbsenceProof(proofStart, proofEnd int, proofNodes [][]byte, leafHash []b
...
 // the provided data in the tree. Additionally, it verifies that the namespace
 // is complete and no leaf of that namespace was left out in the proof.
-func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, data [][]byte, root []byte) bool {
+// leafs contain leaves within the nID
+func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, leaves [][]byte, root []byte) bool {
+	// TODO [Me] what is this check for?
 	nth := NewNmtHasher(h, nID.Size(), proof.isMaxNamespaceIDIgnored)
 	min := namespace.ID(MinNamespace(root, nID.Size()))
 	max := namespace.ID(MaxNamespace(root, nID.Size()))
+	// TODO [Me] this never happens, the min and max are exactly nID.Size() bytes
```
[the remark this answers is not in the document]
staheri14: > The passed-in nID could be any length though

Right, however, just to clarify, what I meant is that the `MinNamespace(root, nID.Size())` and `MaxNamespace(root, nID.Size())` functions are designed to always return a namespace ID with a size of `nID.Size()`. 

https://github.com/celestiaorg/nmt/blob/b04eea51d193f2fa33af6c017979a1516b6865ef/nmt.go#L425

nmt.go
staheri14: Noted! thanks.

nmt.go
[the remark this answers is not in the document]
staheri14: Update: I just double checked the code, and figured these two fields represent the range of the namespace IDs of the data stored by a neamspace tree, and they are accessed in here:
https://github.com/celestiaorg/nmt/blob/ac87f1cd1331b2a6b72b14ab427cffeddccaa154/nmt.go#L174
This means, as it is, we cannot remove these two fields. 
However, we could retrieve the same information i.e., the namespaceID range by looking at the tree [`rawRoot`](https://github.com/celestiaorg/nmt/blob/ac87f1cd1331b2a6b72b14ab427cffeddccaa154/nmt.go#L95) field (as it is supposed to embody the max and min namespace IDs of the entire tree). However, that requires a bit of refactoring so that the root gets calculated after each `Push` operation (Which is not currently the case).

proof.go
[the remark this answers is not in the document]
liamsi: I see. Yes, you are right.

nmt.go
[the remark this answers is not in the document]
liamsi: Yes, removing these fields will need some refactoring. I did not think it through beyond the fact that the info about min/max namespaces also materializes in the root.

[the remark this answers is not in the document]
liamsi: > I see, then, does it mean that the minID is actually equal to 0xFF which corresponds to the Max namespace ID?

No, this is only to initialize the minID such that the if-block yields true for any namespace smaller than the max one (and hence the minID gets updated): https://github.com/celestiaorg/nmt/blob/ac87f1cd1331b2a6b72b14ab427cffeddccaa154/nmt.go#L174

nmt.go
```diff
@@ -142,71 +149,119 @@ func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
...
 	isMaxNsIgnored := n.treeHasher.IsMaxNamespaceIDIgnored()
-	// In the cases (nID < minNID) or (maxNID < nID),
-	// return empty range and no proof:
-	if nID.Less(n.minNID) || n.maxNID.Less(nID) {
+	// case 1)
+	// In the cases (n.nID < minNID) or (n.maxNID < nID),
+	// return empty range and no proof
+	if nID.Less(n.minNID) || n.maxNID.Less(nID) { // TODO [Me] we could move this entire if block inside the `foundInRange` function
+		// TODO [Me] Shouldn't we instead return the first or the last node in the tree as the exclusion proof?
+		// TODO [Me] although I think this current logic is based on the premise that the root of the tree is trusted
```
liamsi, line 176: That is a good question. I thought the root (which commits/includes the min/max nids of the tree) is proof enough that the tree does not contain the passed-in `nID`. Is there a way to construct a tree/root such that the root is not sufficient proof (that the namespace is not included)? I do not think so but maybe I am missing sth. These are exactly the edge-cases that might require further thoughts.

nmt.go
```diff
@@ -142,71 +149,119 @@ func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
...
+	// find the range of indices of leaves with the given nID
 	found, proofStart, proofEnd := n.foundInRange(nID)
+
+	// case 2)
 	if !found {
 		// To generate a proof for an absence we calculate the
 		// position of the leaf that is in the place of where
 		// the namespace would be in:
-		proofStart = n.calculateAbsenceIndex(nID)
+		proofStart = n.calculateAbsenceIndex(nID) // TODO [Me] this could simply return a range, to avoid the line below
```
liamsi, line 188: 👍🏼

nmt.go
```diff
@@ -142,71 +149,119 @@ func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
...
+
 	return NewAbsenceProof(proofStart, proofEnd, proof, n.leafHashes[proofStart], isMaxNsIgnored), nil
 }
 
+// buildRangeProof returns the nodes (as byte slices) in the range proof of the supplied range i.e.,
+// [proofStart, proofEnd) where proofEnd is non-inclusive.
+// The nodes are ordered according to in order traversal of the namespaced tree. (TODO [Me] I need to double-check the order)
 func (n *NamespacedMerkleTree) buildRangeProof(proofStart, proofEnd int) [][]byte {
-	proof := [][]byte{}
+	// TODO a more secure way would be to make it slice of fixed size arrays i.e., the hash output size
```
liamsi, line 215: In that case, would we also return a slice of fixed size arrays? What if someone wants to use a different hash function (with a different output size)?

nmt.go
```diff
@@ -142,71 +149,119 @@ func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
...
 	var recurse func(start, end int, includeNode bool) []byte
+
+	// start, end are indices of leaves in the tree hence they should be within the size of the tree i.e.,
+	// less than or equal to the len(n.leaves)
+	// this recursive algorithm is inspired by RFC6962 https://www.rfc-editor.org/rfc/rfc6962#section-2.1
+	// includeNode indicates whether the hash of the current subtree (covering the supplied range i.e., [start, end)) or
+	// one of its constituent subtrees should be part of the proof
 	recurse = func(start, end int, includeNode bool) []byte {
-		if start >= len(n.leafHashes) {
+		if start >= len(n.leafHashes) { // TODO [Me] why against leafHashes? and not leaves?
```
liamsi, line 225: No particular reason. The lengths of both should be the same.

nmt.go
```diff
@@ -142,71 +149,119 @@ func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
...
+			// TODO [Me] Why there is no check for the end index?
```
liamsi, line 226: I _think_ it's not necessary by construction. Maybe the same actually holds for start as well though. Would need to verify that start+k never can exceed len(leaves):
https://github.com/celestiaorg/nmt/blob/29cca3c0d30b05545bf37c09fa19dc070289610a/nmt.go#L211

nmt.go
```diff
@@ -142,71 +149,119 @@ func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
...
-		// recursively get left and right subtree
+
+		// newIncludeNode indicates whether one of the subtrees of the current subtree [start, end)
+		// may have an overlap with the queried proof range i.e., [proofStart, proofEnd)
 		newIncludeNode := includeNode
-		if (end <= proofStart || start >= proofEnd) && includeNode {
+		// check whether the subtree representing the [start, end) range of leaves has overlap with the
+		// queried proof range i.e., [proofStart, proofEnd)
+		// if not
+		if (end <= proofStart || start >= proofEnd) && includeNode { //TODO [Me] the `&& includeNode` seems ineffective and unnecessary
```
liamsi, line 251: Not sure I understand this 🤔

proof.go
```diff
@@ -113,25 +116,27 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, data [][]byte,
 		}
 		return false
 	}
-	gotLeafHashes := make([][]byte, 0, len(data))
+	// TODO [Me] we could also check whether the number of leaves match the proof range end-start and make an early return if not
```
liamsi, line 119: 👍🏼

proof.go
```diff
@@ -113,25 +116,27 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, data [][]byte,
...
 		hashLeafFunc := nth.HashLeaf
-		for _, gotLeaf := range data {
+		for _, gotLeaf := range leaves {
+			// TODO [Me] can be converted to something like isNameSpaceIDPrefixed()
 			if len(gotLeaf) < int(nIDLen) {
 				// conflicting namespace sizes
 				return false
 			}
-			gotLeafNid := namespace.ID(gotLeaf[:nIDLen])
+			gotLeafNid := namespace.ID(gotLeaf[:nIDLen]) // TODO [Me] a helper function
```
liamsi, line 134: 👍🏼 everything that improves readability (without impacting performance much) should be done.

proof.go
```diff
@@ -142,8 +147,11 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, data [][]byte,
 	return proof.verifyLeafHashes(nth, true, nID, gotLeafHashes, root)
 }
 
-func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID namespace.ID, gotLeafHashes [][]byte, root []byte) bool {
+// verifyLeafHashes checks whether all the leaves matching the namespace ID nID are covered by the proof if verifyCompleteness is set to true
+func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID namespace.ID, leafHashes [][]byte, root []byte) bool {
 	var leafIndex uint64
+	// TODO [Me] Why called leftSubtrees?
```
liamsi, line 153: Not sure what would be a better name. LeftSubtreeRoots maybe?

proof.go
```diff
@@ -162,7 +170,7 @@ func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID na
 				return false
 			}
 		}
-		// rightSubtrees only contains the subtrees after [0, r.Start)
+		// rightSubtrees only contains the subtrees after [0, r.end)
```
liamsi, line 173: oof, great find. Maybe we should use capitalisation here consistently? `[0, r.End)`

nmt.go
staheri14: My suggestion is about a small simplification on the if statement.
What I meant is that  if we remove the condition on the `includeNode` from the `if` statement, we could get the same result, i.e.,
```
newIncludeNode := includeNode
if (end <= proofStart || start >= proofEnd) {
			newIncludeNode = false
		}
``` 
The reason is that:
In the original version, If `includeNode` is `false`,  the `if` block is never reached, hence the `newIncludeNode` contains the `includeNode` value i.e., `false` (the same result is yield in the new if statement)
In the original version, if the `includeNode` is `true`, and if the first condition is `true`  i.e., `(end <= proofStart || start >= proofEnd)` then we enter the if block and toggle the  `newIncludeNode` to `false` (which is the same result we obtain from the new if statement).

proof.go
staheri14: My question was originally about why the the data populated inside `leftSubtrees` represent the left subtrees. Later, by examining the code, I figured why. I should have removed this comment. 
Nevertheless, I liked your suggested name i.e., `LeftSubtreeRoots`, it is more indicative of the content of the variable.

nmt.go
staheri14: If all the querying nodes have access to the correct NMT roots, then there should be no problem hence no exclusion proof is required (I will still think about it to see if any corner cases may happen). Thanks for your answer!

nmt.go
staheri14: If we expect varying namespace ID sizes, then nodes cannot be represented by a fixed size array.

nmt.go
staheri14: Thanks for clarification, right, `end` can actually exceed the `len(n.leafHashes)` because we call `recurse` on the full tree size i.e., https://github.com/celestiaorg/nmt/blob/ac87f1cd1331b2a6b72b14ab427cffeddccaa154/nmt.go#L288

staheri14:
@liamsi I have added a few other questions and suggestions, would appreciate your review and thoughts on them.

staheri14:
Closing the PR since it has fulfilled its intended purpose, and several of the questions raised in this conversation have been converted into corresponding GitHub (GH) issues or addressed in separate PRs. If there are any remaining matters that need attention, I will address them by creating additional GH issues.

PR #100, opened 2023-02-14
feat!: increase default namespace id size to 16 bytes

rootulp:
Part of https://github.com/celestiaorg/celestia-app/issues/1308

Blocked on an ADR in the celestia-app repo that explores this decision


rootulp:
No longer needed

PR #101, opened 2023-02-15
chore: NMT specifications

staheri14:
## Overview
This pull request aligns with the https://github.com/celestiaorg/celestia-app/issues/1296 EPIC and introduces an NMT specification that includes a description of NMT as a data structure, an explanation of the NMT library, and a sample code. In terms of description of the NMT library, the focus was on the portions that are used in the NMT wrapper to establish the foundation for the NMT wrapper specification. The NMT methods are illustrated through a running code example that was originally included in the Readme file of the NMT repository (with a minor change in namespace ID of one of the leaves in order to be able to provide an example of absence proof). This example has been expanded to provide a more detailed description of each method call, an interpretation of the parameters, and its relationship to the NMT high-level logic. The The corresponding NMT for the code example is also visualized to facilitate a better understanding and examination of the expected behavior of the library.

### DIFF
--- a/spec/nmt.md
+++ b/spec/nmt.md
@@ -9,8 +9,9 @@ This specification explains the NMT data structure and provides an overview of i
 
 # NMT Data Structure
 
-Namespaced Merkle Tree, at the core is a normal Merkle tree that employs a modified hash function, namely a [namespaced hash](#namespaced-hash) to ensure each node in the tree encompasses the range of namespaces of its descendants' messages. 
-The messages stored in the NMT leaves are arranged in ascending order based on their namespace IDs, which must be of a fixed and known size and have the format `<NsID>||<Message Data>`. 
+Namespaced Merkle Tree, at its core, is a normal Merkle tree that employs a modified hash function, namely a [namespaced hash](#namespaced-hash) to ensure each node in the tree encompasses the range of namespaces of its descendants' messages. 
+Messages stored in the NMT leaves are all namespace-prefixed with the format `<NsID>||<Message Data>` and arranged in ascending order based on their namespace IDs.
+All namespace identifiers have a fixed and known size. 
 
 
 ## Namespaced Hash
--- a/spec/nmt.md
+++ b/spec/nmt.md
@@ -65,8 +65,8 @@ For the sake of brevity, we have only included the first 7 hex digits of SHA256 
 00 6c6561665f30      00 6c6561665f31    01 6c6561665f32      03 6c6561665f33  Namespaced Data Items 
 
         0                   1                  2                  3           Data Items/Leaves Indices
-```
-                                        Figure 1.
+``` 
+Figure 1.
 
 ## Namespace Proof
 
--- a/spec/nmt.md
+++ b/spec/nmt.md
@@ -43,7 +43,8 @@ The inclusion of the `0x01` value in the hash calculation serves as a non-leaf p
 In an NMT data structure, the `minNs` and `maxNs` values of the root node denote the minimum and maximum namespace IDs, respectively, of all messages within the tree.
 
 
-An example of an NMT is shown in the figure below which utilizes SHA256 as the underlying hash function and namespace ID size of `1` byte. The code snippets necessary to create this tree are provided in [NMT Library](#nmt-library) section, and the data items and tree nodes are represented as hex strings. 
+An example of an NMT is shown in the figure below which utilizes SHA256 as the underlying hash function and namespace ID size of `1` byte. 
+The code snippets necessary to create this tree are provided in [NMT Library](#nmt-library) section, and the data items and tree nodes are represented as hex strings. 
 For the sake of brevity, we have only included the first 7 hex digits of SHA256 for each namespace hash.
 
 ```markdown
--- a/spec/nmt.md
+++ b/spec/nmt.md
@@ -112,7 +112,7 @@ In Figure 1, a query for `nID=6` would be responded by an empty proof as it fall
 
 ### Verification of NMT Inclusion Proof
 
-An NMT inclusion proof is deemed valid if it meets the followings:
+An NMT inclusion proof is deemed valid if it meets the following:
 - The namespace ID of the leaves in the returned range `[start, end)` match the queried `nID`.
 - **Inclusion**: The supplied Merkle proof for the range `[start, end)` is valid for the given tree root `T`. 
 - **Completeness**: There are no other leaves matching `nID` that do not belong to the returned range `[start, end)`.
--- a/spec/nmt.md
+++ b/spec/nmt.md
@@ -16,14 +16,13 @@ All namespace identifiers have a fixed and known size.
 
 ## Namespaced Hash
 
-NMT utilizes a namespaced hash function, which in addition to the normal digest calculation, it returns the range of namespace IDs covered by a node's left and right children (in case of non-leaf nodes), or the namespace id range of the namespaced message (in case of leaf nodes).
-The hash output is formatted as  `minNs||maxNs||h(.)`, where `minNs` is the lowest namespace identifier among all the node's descendants, `maxNs` is the highest, and `h` represents the hash digest (e.g., SHA256).
+NMT utilizes a namespaced hash function i.e., `NsH()`, which in addition to the normal digest calculation, it returns the range of namespace IDs covered by a node's children. The hash output is formatted as  `minNs||maxNs||h(.)`, where `minNs` is the lowest namespace identifier among all the node's descendants, `maxNs` is the highest, and `h` represents the hash digest (e.g., SHA256).
 
-**Leaf Nodes**: Each leaf in the tree represents the namespaced hash of a namespaced message `d` with the following format `d = <NsID>||<Message Data>`. 
+**Leaf Nodes**: Each leaf in the tree represents the namespaced hash of a namespaced message `d = <NsID>||<Message Data>`. 
 The hash is computed as follows:
 
 ```go
-namespacedHash(d) = d.NsID || d.NsID || h(0x00, d)
+NsH(d) = d.NsID || d.NsID || h(0x00, d)
 ```
 The inclusion of the `0x00` value in the hash calculation serves as a leaf prefix and is done to conform to [RFC 6962](https://www.rfc-editor.org/rfc/rfc6962#section-2.1).
 
@@ -36,7 +35,7 @@ The inclusion of the `0x00` value in the hash calculation serves as a leaf prefi
 the namespaced hash is calculated as
 
 ```go
-namespacedHash(n) = min(l.minNs, r.minNs) || max(l.maxNs, r.maxNs) || h(0x01, l, r)
+NsH(n) = min(l.minNs, r.minNs) || max(l.maxNs, r.maxNs) || h(0x01, l, r)
 ```
 
 The inclusion of the `0x01` value in the hash calculation serves as a non-leaf prefix and is done to conform to [RFC 6962](https://www.rfc-editor.org/rfc/rfc6962#section-2.1).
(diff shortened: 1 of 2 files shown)


spec/nmt.md
```diff
@@ -0,0 +1,303 @@
+# Introduction
+
+Namespaced Merkle Tree, NMT for short, is one of the core components of Celestia blockchain.
+Transactions in Celestia are associated with a namespace ID which signifies the application they belong to.  
+Nodes interested in a specific application only need to download transactions of a certain namespace ID. 
+The Namespaced  Merkle Tree (NMT) was introduced in the [LazyLEdger article](https://arxiv.org/abs/1905.09274) to organize transactions in Celestia blocks based on their namespace IDs.
```
liamsi, line 6:
```suggestion
The Namespaced  Merkle Tree (NMT) was introduced in the [LazyLedger article](https://arxiv.org/abs/1905.09274) to organize transactions in Celestia blocks based on their namespace IDs. 
```

liamsi (commented):
This looks great 👏🏼 Thanks!

A few comments (did just a quick first pass; will do a more thorough review in the next 2 days):

 - We should consider separating the high-level spec from the library documentation (both are valuable to have and IMO the # NMT Library section could live in a separate document which could be linked from the main readme)
 - Suggestion: should we add/append some test-vectors to the spec so implementers can check against these
 - Simple single-leaf inclusion proofs or range-proofs that concern only a subset of a namespace are not covered. Q: Is this presumably because these are the same as for rfc6962? Does it still make sense to include these?

spec/nmt.md
staheri14: > We should consider separating the high-level spec from the library documentation (both are valuable to have and IMO the # NMT Library section could live in a separate document which could be linked from the main readme)

Separation of these two makes sense to me as well, I was also initially thinking to do the same. Do we want to put the file for the NMT Library description under the spec repo? or a separate repo?

> Suggestion: should we add/append some test-vectors to the spec so implementers can check against these

Could you provide more information on the specific type of test vectors you're suggesting? Are you proposing to include these test vectors in the high-level specification or the NMT library section? Figure 1 already shows an NMT that one could create by following the code example in the  NMT library section (the exact hash values (the first 7 hex digits) are provided in that figure). Additionally,  I have also provided some sample namespace queries and proof results in the spec section. I can certainly provide more elaborate examples, please let me know.

> Simple single-leaf inclusion proofs or range-proofs that concern only a subset of a namespace are not covered. Q: Is this presumably because these are the same as for rfc6962? Does it still make sense to include these?

The logic of a normal Merkle range proof and verification is actually already covered in the description of namespace proof generation and verification. That is, each namespace proof consists of two parts: 1- finding the range of leaves matching the queried namespace 2- calculating the Merkle range proof.  Similarly, the namespace proof verification involves Merkle range proof verification. However, to make it more clear for the readers, I can add an extra subsection dedicated to the Merkle proofs only. Please let me know your thoughts.

spec/nmt.md
```diff
@@ -0,0 +1,303 @@
+# Introduction
+
+Namespaced Merkle Tree, NMT for short, is one of the core components of Celestia blockchain.
+Transactions in Celestia are associated with a namespace ID which signifies the application they belong to.  
+Nodes interested in a specific application only need to download transactions of a certain namespace ID. 
+The Namespaced  Merkle Tree (NMT) was introduced in the [LazyLedger article](https://arxiv.org/abs/1905.09274) to organize transactions in Celestia blocks based on their namespace IDs. 
+The NMT allows for efficient and verifiable queries of application-specific transactions by accessing only the block header, which contains the NMT root.
+This specification explains the NMT data structure and provides an overview of its current implementation in the repository.
```
evan-forbes, line 8: [non-blocking question]

are we purposefully using a new line on line 4? if so, do we want to add another new line for formatting?

spec/nmt.md
```diff
@@ -0,0 +1,303 @@
...
+In Figure 1, a query for `nID=6` would be responded by an empty proof as it falls outside the root's namsespace range.
+
+## Namespace Proof verification
+
+### Verification of NMT Inclusion Proof
+
+An NMT inclusion proof is deemed valid if it meets the followings:
+- The namespace ID of the leaves in the returned range `[start, end)` match the queried `nID`.
+- **Inclusion**: The supplied Merkle proof for the range `[start, end)` is valid for the given tree root `T`. 
+- **Completeness**: There are no other leaves matching `nID` that do not belong to the returned range `[start, end)`.
```
evan-forbes, line 119: nice, this terminology makes a lot of sense to me

evan-forbes (approved):
really great work! thorough && concise!

no blocking comments from my end, all LGTM

spec/nmt.md
```diff
@@ -0,0 +1,303 @@
...
+// compute the root
+root := tree.Root() 
+```
+In the provided code example, the root would be `00 03 b1c2cc5` (as also illustrated in Figure 1).
+
+The minimum and maximum namespace IDs of the tree root can be obtained through the following methods:
+
+```go
+minNS := nmt.MinNamespace(root, tree.NamespaceSize())
+maxNS := nmt.MaxNamespace(root, tree.NamespaceSize())
```
rootulp, line 227: [no change needed] I wonder if these methods should be added because they seem slightly more conveinet:

```go
minNS := tree.MinNamespace() 
maxNS := tree.MaxNamespace() 
```

given the tree struct already has exported methods for `tree.Root()` and `tree.NamespaceSize()`, this seems feasible.

spec/nmt.md
```diff
@@ -0,0 +1,303 @@
...
+```
+ The `minNs` and `maxNs` are equal to `00` and `03` in the supplied example.
+
+## Generate Namespace Proof
+
+The `ProveNamespace` can be used to generate a namespace proof for a specific namespace ID.
```
rootulp, line 233: [nit]

```suggestion
The `ProveNamespace` method can be used to generate a namespace proof for a specific namespace ID.
```

spec/nmt.md
```diff
@@ -0,0 +1,303 @@
...
+```go
+func (n *NamespacedMerkleTree) ProveNamespace(nID namespace.ID) (Proof, error)
+```
+For example:
+```go
+nID := namespace.ID{0}
+proof, err := tree.ProveNamespace(nID)
+    if err != nil {
+      panic("unexpected error")
+}
```
rootulp, line 244: [nit] indentation appears off

```suggestion
if err != nil {
	panic("unexpected error")
}
```

spec/nmt.md
staheri14: Good catch! the newline is not intentional on that line, fixed it now https://github.com/celestiaorg/nmt/pull/101/commits/440b7623b1fcd5cd93aa57c6dbcadc714abf0207

spec/nmt.md
```diff
@@ -0,0 +1,303 @@
...
+              /            \                    /               \
+            NsH()          NsH()              NsH()             NsH() 
+            /                \                /                   \
+    00 00 5fa0c9c       00 00 52385a0    01 01 71ca46a       03 03 b4a2792    Leaf Nodes
+        |                   |                 |                   |
+      NsH()               NsH()              NsH()               NsH()
+        |                   |                 |                   |
+00 6c6561665f30      00 6c6561665f31    01 6c6561665f32      03 6c6561665f33  Namespaced Data Items 
+
+        0                   1                  2                  3           Data Items/Leaves Indices
```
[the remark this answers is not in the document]
staheri14: Thanks for the suggestion. I initially considered using only the term `leaf indices`, but I realized that it could cause confusion about whether the row above is referring to leaves or data items. Therefore, I decided to include both `Data Items/Leaves Indices` to clarify the meaning.  What do you think?

spec/nmt.md
```diff
@@ -0,0 +1,303 @@
...
+### Namespace Absence Proof
+
+If the namespace ID being queried falls within the range of namespace IDs in the tree root, but there is no corresponding message, an absence proof will be generated. 
+An absence proof asserts that no message in the tree matches the queried namespace ID `nID`.
+
+The absence proof consists of:
+1) The index of a leaf of the tree that 
+   1) its namespace ID is the smallest namespace ID larger than `nID` and
+   2) the namespace ID of the leaf to the left of it  is smaller than `nID` 
+   3) the namespace ID of the leaf to the right of it is larger than  `nID`.
```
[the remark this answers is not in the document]
staheri14: If we add an `if any` condition to the third one that will capture the edge case you mentioned i.e.,
`   3) the namespace ID of the leaf to the right of it, if any, is larger than  nID`.
However, I agree that the the first two imply the third condition, thus going to remove it.

spec/nmt.md
```diff
@@ -0,0 +1,303 @@
...
+Note that the proof only requires the hash of the leaf, not its underlying message. 
+This is because the aim of the proof is to demonstrate the absence of `nID`.
+
+In Figure 1, if we query for `nID = 2`, we will receive an absence proof since there is no matching item for it. 
+The index of the leaf included in the proof will be `3`, which corresponds to the node `03 03 b4a2792`. 
+The Merkle inclusion proof for this leaf consists of the following nodes, in the order they appear in an in-order traversal of the tree: `00 00 ead8d25` and `01 01 71ca46a`.
+
+
+### Namespace Empty Proof
+If the requested namespace ID falls outside the namespace range represented by the tree root `T`, the query will be resolved with an empty namespace proof.
```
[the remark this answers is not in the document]
staheri14: Good question, yes, this is actually the case in the current implementation. 
I am not quite sure about its use-case though. In fact, I originally suggested to provide an exclusion proof for such queries instead of returning an empty proof. Here is a link to this conversation in another PR https://github.com/celestiaorg/nmt/pull/94/files#r1098349894. Once the final decision is made, I'll update the spec.

spec/nmt.md
staheri14: Sure, noted https://github.com/celestiaorg/nmt/issues/104

staheri14:
Thanks @rootulp for your comments! I addressed them all in the following commits:
https://github.com/celestiaorg/nmt/pull/101/commits/a9d0d5a369528acf639aa67bde16638b2db7bef2
https://github.com/celestiaorg/nmt/pull/101/commits/03a7cdb1018fcaaa6dc3c59e0495634fdac15d0d
https://github.com/celestiaorg/nmt/pull/101/commits/62dfd1094ce506024736373b10abffb726f2242c
https://github.com/celestiaorg/nmt/pull/101/commits/bad1cc08ca3c1940fb85f409dcc719d33f56a38f

spec/nmt.md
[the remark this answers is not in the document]
staheri14: Update: To be more exact, my suggestion is part of the following comment in that said PR, please see https://github.com/celestiaorg/nmt/blob/43a8d1e6271aff69cf2bf1cb861f5ff79c0984f9/nmt.go#L180

evan-forbes (approved):

spec/nmt.md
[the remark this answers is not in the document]
rootulp: thanks for the link! I think we can resolve this convo because you're already tracking the question in https://github.com/celestiaorg/nmt/pull/94/files#r1098349894

spec/nmt.md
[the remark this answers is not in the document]
rootulp: Since the row above has the label Namespaced Data Items, I got confused by seeing Data Items again on this row. I think Leaf Indices is clearer but defer to your judgement.

spec/nmt.md
```diff
@@ -0,0 +1,303 @@
+# Introduction
+
+Namespaced Merkle Tree, NMT for short, is one of the core components of Celestia blockchain.
+Transactions in Celestia are associated with a namespace ID which signifies the application they belong to.  
+Nodes interested in a specific application only need to download transactions of a certain namespace ID. 
+The Namespaced  Merkle Tree (NMT) was introduced in the [LazyLedger article](https://arxiv.org/abs/1905.09274) to organize transactions in Celestia blocks based on their namespace IDs. 
+The NMT allows for efficient and verifiable queries of application-specific transactions by accessing only the block header, which contains the NMT root.
+This specification explains the NMT data structure and provides an overview of its current implementation in the repository.
+
+# NMT Data Structure
```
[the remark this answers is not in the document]
rootulp: Created https://github.com/celestiaorg/nmt/issues/106 so resolving this conversation

spec/nmt.md
liamsi: > Separation of these two makes sense to me as well, I was also initially thinking to do the same. Do we want to put the file for the NMT Library description under the spec directory? or a separate directory?

I think the following would make sense
/
|
+ docs
     |
    + specs
    
And I think the library description can go under docs and the specs under specs. No need to tackle this here. We can do this in followup PRs.

> The logic of a normal Merkle range proof and verification is actually already covered in the description of namespace proof generation and verification. 

Makes sense!

> Could you provide more information on the specific type of test vectors you're suggesting?

I think it would be good to have an appendix with realistic testvectors that implementers of nmt (specifically for celestia) could test against. Ideally, these should have the same params as celestia then. We can generate these once we have full confidence in the implementation.

liamsi (approved):
Also good to go from my side.

spec/nmt.md
staheri14: > I think it would be good to have an appendix with realistic testvectors that implementers of nmt (specifically for celestia) could test against. Ideally, these should have the same params as celestia then. We can generate these once we have full confidence in the implementation.

Thank you for the clarification. I have created a tracking issue (https://github.com/celestiaorg/nmt/issues/111) to address this once the implementation becomes stable.

PR #102, opened 2023-02-16
feat: verify the range of the namespace IDs of the children in the HashNode

### LINKED ISSUE #95: feat: verify the range of the namespace IDs of left and right nodes in the Hasher
# Problem
The [ `Hasher`](https://github.com/celestiaorg/nmt/blob/b04eea51d193f2fa33af6c017979a1516b6865ef/hasher.go#L155) implementation does not verify the range of namespace IDs in the left and right nodes with respect to each other, which is crucial for proper ordering of leaves based on their namespace IDs and for detecting invalidly constructed trees and proofs.

Quoting from Lazy Ledger paper:
> An adversarial consensus node may attempt to produce a block that contains a Merkle tree with children that are not ordered correctly. To prevent this, we can set a condition in nsHash such that there is no valid hash when leftMaxNs ≥ rightMinNs, and thus there would be no valid Merkle root for incorrectly ordered children

The issue was identified as part of https://github.com/celestiaorg/celestia-app/issues/1296
# Acceptance Criteria
- [ ] To modify the [`Hasher`](https://github.com/celestiaorg/nmt/blob/b04eea51d193f2fa33af6c017979a1516b6865ef/hasher.go#L155) implementation by incorporating this additional namespace ID range check for the supplied children nodes. Possibly, the `Hasher` can return an error for invalid inputs. Subsequent refactor needs to take place to properly consume this change.

staheri14:
## Overview
This PR closes #95. 
Inline with https://github.com/celestiaorg/celestia-app/issues/1296.

The implementation of this pull request does not exactly follow the original suggestion in the LazyLedge paper. Specifically, it does not treat `leftMaxNs = rightMinNs` as a violation of leaf ordering. The reason for this is that a namespace may span across multiple subtrees.
cc: @liamsi  

Quoting from the article:
> An adversarial consensus node may attempt to produce a block that contains a Merkle tree with children that are not ordered correctly. To prevent this, we can set a condition in nsHash such that there is no valid hash when **leftMaxNs ≥ rightMinNs**, and thus there would be no valid Merkle root for incorrectly ordered children

## Checklist


hasher_test.go
```diff
@@ -86,14 +86,6 @@ func Test_namespacedTreeHasher_HashNode(t *testing.T) {
 				sum(crypto.SHA256, []byte{NodePrefix}, []byte{0, 0, 0, 0}, []byte{0, 0, 1, 1})...,
 			),
 		},
-		{
-			"leftmin==rightmin && leftmax>rightmax", 2,
-			children{[]byte{0, 0, 1, 1}, []byte{0, 0, 0, 1}},
-			append(
-				[]byte{0, 0, 1, 1},
-				sum(crypto.SHA256, []byte{NodePrefix}, []byte{0, 0, 1, 1}, []byte{0, 0, 0, 1})...,
```
staheri14, line 94: This is an invalid state given the newly added logic i.e., the max namespace of the right child cannot be greater than the min namespace of the right child.

hasher.go
```diff
@@ -158,6 +160,13 @@ func (n *Hasher) HashNode(left, right []byte) []byte {
 	leftMinNs, leftMaxNs := left[:n.NamespaceLen], left[n.NamespaceLen:flagLen]
 	rightMinNs, rightMaxNs := right[:n.NamespaceLen], right[n.NamespaceLen:flagLen]
 
+	// check the namespace range of the left and right children
+	rightMinNID := namespace.ID(rightMinNs)
+	leftMaxNID := namespace.ID(leftMaxNs)
+	if rightMinNID.Less(leftMaxNID) {
+		panic("nodes are out of order: the maximum namespace of the left child is greater than the min namespace of the right child")
```
evan-forbes, line 167: would panicking here mean that someone could send a light node an invalid proof or sample and their node would crash?

evan-forbes (commented):
one question, otherwise LGTM!

hasher.go
staheri14: As you mentioned, the `HashNode` method may only fail in the event of an attack on the system. However, it is important to note that the method was previously susceptible to panic, even prior to this pull request. For example, panic can occur while extracting the namespace identifier of the left or right children if they are not namespaced (i.e., if their length is less than NamespaceLen). It is not possible to return an error from this method without violating the interfaces it is intended to fulfill, namely `NodeHasher` and `hash`.
At this time, it may be better to allow the light client to crash rather than enter an invalid state.

I will investigate the implementation of proper error handling in the `HashNode` method, including handling other potential errors beyond just the namespace ID range, in a follow-up pull request. As this has a broad scope, it requires careful attention and consideration during implementation.

liamsi (commented):
>  Specifically, it does not treat leftMaxNs = rightMinNs as a violation of leaf ordering. The reason for this is that a namespace may span across multiple subtrees.

I see. That makes sense. Maybe @musalbas wants to update that case in the paper too.

hasher.go
liamsi: Hmm, I think when this was implemented, we might have panic'ed bc it was supposed to be impossible to construct an invalid tree. But it is good thinking that someone might do so to crash light nodes and I do think we should error instead.

liamsi: For the sake of getting this PR merged: can we add this property/panic to the godoc? So the caller has a hint that verification has to happen before to not cause unexpected panics?

staheri14: > For the sake of getting this PR merged: can we add this property/panic to the godoc? So the caller has a hint that verification has to happen before to not cause unexpected panics?

Sure! please check the following commit https://github.com/celestiaorg/nmt/pull/102/commits/e7415b7956421365459b0e3b37340c9bc4f834fc

evan-forbes (approved):
:+1: :+1: 

we might also want to handle #109 before we cut a new release to avoid any unintentional panics

staheri14:
> 👍 👍
> 
> we might also want to handle #109 before we cut a new release to avoid any unintentional panics

Sure @evan-forbes, I am on it!

staheri14:
@liamsi I'd appreciate if you could please take another look and let me know if further updates are required on the PR before merging. Thanks!

liamsi (approved):
🙏🏼

Issue #104, opened 2023-02-17
feat: expose methods to get min and max namespace IDs of an NMT


staheri14:
# Problem
Currently, there is no exposed method in the nmt API to get the min and max namespace IDs of an NMT. To do so, one needs to compute the root and then make another function call to spot the min and max namespace IDs encoded in the root. 

# Acceptance Criteria
- [ ] Augment the current NMT API with `getMinNamespace` and `getMaxNamespace` as methods of the `NamespaceMerkleTree` struct.  Proper unit tests should be included as well.

PR #105, opened 2023-02-17
feat: implement nmt methods Min/MaxNamespace

rootulp:
Closes https://github.com/celestiaorg/nmt/issues/104

### DIFF
--- a/nmt_test.go
+++ b/nmt_test.go
@@ -14,6 +14,7 @@ import (
 	"testing"
 
 	"github.com/celestiaorg/nmt/namespace"
+	"github.com/stretchr/testify/assert"
 )
 
 type namespaceDataPair struct {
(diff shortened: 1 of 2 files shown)


nmt_test.go
```diff
@@ -773,3 +774,97 @@ func generateRandNamespacedRawData(total int, nidSize int, leafSize int) [][]byt
...
+		{
+			name: "example tree with eight leaves",
+			tree: exampleTreeWithEightLeaves(),
+			want: namespace.ID{8, 8},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			got := tc.tree.MaxNamespace()
```
staheri14, line 826: [optional] It's totally up to you, but I wanted to suggest that combining these two tests i.e., `TestMaxNamespace` and `TestMinNamespace` might be a good idea. The test prep for both of them is pretty much identical, which means we would only need to build one set of NMTs.  Also and the logic of the two methods are very similar. This could potentially give a slightly higher performance in our tests.

staheri14 (approved):
I noticed that the PR is still in draft mode, but I just wanted to leave my first review since I received the request. If you're still working on it and plan to make further updates, just let me know and I'll take another look later. But for now, everything looks good to me - LGTM!

rootulp:
Sorry this PR was still in draft b/c I was waiting for CI checks to pass before marking as ready for review.

nmt_test.go
rootulp: Yea good point. I almost did this originally but I like when each method (e.g. `Method`) has it's own unit test (e.g. `TestMethod`) so that it's easy to navigate from method to unit test and see what the expected behavior is like. Since that isn't the case for other unit test names in this file, I'll consolidate.

rootulp: Resolved in https://github.com/celestiaorg/nmt/pull/105/commits/893ff0375a789dad850c9714b0b6a7993695e5ad

Issue #99, opened 2023-02-14
fix: Resolving Inconsistency in the ProveNamespace Signature for Error Handling


staheri14:
# Problem
Identified as part of this EPIC https://github.com/celestiaorg/celestia-app/issues/1296.
This issue is to investigate and address the inconsistency in the `ProveNamespace` signature, where it indicates that the method should return an `error` but in its current implementation, the error value is always `nil`. 

# Acceptance Criteria
Identify the cases where the method should return a proper error or remove the error from the method signature if no error conditions are perceived. Refactor the code, if necessary.

staheri14:
This issue has been addressed in https://github.com/celestiaorg/nmt/pull/136

Issue #97, opened 2023-02-13
Defining the Verification Process for NMT Proofs Generated from Non-Lexicographically Ordered Trees


staheri14:
# Problem
The issue at hand pertains to the generation of NMTs from sets of leaves that are not ordered lexicographically based on their namespaces. The behavior of NMT proofs generated from such malformed trees is unclear and needs to be addressed. The main objectives of this issue are to:

- Determine whether the current NMT proof verification system is able to detect out-of-order leaves and respond accordingly.
- If the current system is not capable, then develop and implement measures to invalidate NMT proofs generated from malformed NMTs.

evan-forbes:
just leaving a thought here: I think we'll have the functionality to verify an out of order nmt inclusion proof, but also have some way of notifying the user that one or more nodes was out of lexicographical order. This way, we could prove that an invalid order was committed to.

staheri14:
Per conversation with @evan-forbes, I'd like to provide more detail on the attack scenario that prompted the opening of this issue:

Malicious consensus nodes could produce and commit a block, denoted as `B`, in which the shares are not ordered lexicographically based on their namespace IDs. A DA (data availability) full node obtains the block header, including the data root denoted as `dataRoot`, and the underlying shares, and will attempt to construct the block. However, if the node detects that the shares used to construct `dataRoot` are unordered, it must provide a proof of bad encoding to the light clients. Specifically, it must indicate that there are out-of-order leaves in the tree represented by `dataRoot`.

The problem is that the current NMT (namespace Merkle tree) implementation does not support this kind of bad-encoding proof. Furthermore, the library does not even allow for the construction of a tree with unordered leaves, since it is protected through the `Push` method, which checks the namespace ID of the inserted data items. Therefore, the issue at hand is to implement new logic in the NMT library that would allow a full node to provide such bad-encoding proof to the light clients.

staheri14:
It might be already covered  by the existing bad encoding fraud proof https://github.com/celestiaorg/celestia-node/blob/main/share/eds/byzantine/bad_encoding.go, we need to check.

PR #113, opened 2023-02-24
feat: implements input validation functions for HashNode and HashLeaf to prevent panics

### LINKED ISSUE #109: feat: Handling Panics in HashNode Method
# Problem
Identified as part of this EPIC: https://github.com/celestiaorg/celestia-app/issues/1296.
The HashNode method of the NMT may potentially cause a panic if it receives invalid or malformed inputs (as mentioned by @evan-forbes in this comment https://github.com/celestiaorg/nmt/pull/102#discussion_r1110338821). For example, if the left or right children do not conform to the namespace hash format, meaning their length is less than the namespace ID size, or if the relation between the left and right children namespace range is invalid. If an invalid node is sent to a light node, it can result in a crash.

It is important to note that returning an error from this method is not possible as it would violate the interface it is intended to fulfill i.e., `NodeHasher` and `Hash`.

# Acceptance Criteria
This issue is to implement proper error handling mechanism  for `HashNode`.

staheri14:
## Overview
An incremental PR toward #109 
**Next**: The validation utility functions developed in this PR must be further integrated into both the tree construction and proof verification process to detect invalid inputs and notify the caller of any errors. This will close #109  and  is in line with the objectives outlined in issues #99 and #97. I will address this in a subsequent pull request.
## Checklist


hasher.go
```diff
@@ -113,42 +117,107 @@ func (n *Hasher) EmptyRoot() []byte {
...
+	return nameSpacedHash
+}
+
+// validateNodeFormat checks whether the supplied node conforms to the
+// namespaced hash format.
+func (n *Hasher) validateNodeFormat(node []byte) (err error) {
+	totalNameSpaceLen := 2 * n.NamespaceLen
+	nodeLen := len(node)
+	if nodeLen < int(totalNameSpaceLen) {
+		return fmt.Errorf("%w: got: %v, want >= %v", ErrMismatchedNamespaceSize, nodeLen, totalNameSpaceLen)
```
rootulp, line 165: [optional] What are your opinions on introducing a new error for this scenario. I ask because it seems possible to encounter this error when there are matching namespace ids but those namespace IDs are too short. Proposal something like:

```diff
var (
	ErrInvalidRange            = errors.New("invalid proof range")
	ErrMismatchedNamespaceSize = errors.New("mismatching namespace sizes")
+	ErrInvalidNodeLen          = errors.New("invalid node length")
	ErrInvalidPushOrder        = errors.New("pushed data has to be lexicographically ordered by namespace IDs")
	noOp                       = func(hash []byte, children ...[]byte) {}
)
```

```suggestion
		return fmt.Errorf("%w: got: %v, want >= %v", ErrInvalidNodeLen, nodeLen, totalNameSpaceLen)
```

For some more context, assume namespace ID size is 8 bytes. If a caller invokes

```go
left := []byte{0, 0, 0, 0}
right := []byte{1, 1, 1, 1}
ValidateNodes(left, right)
```

then the error message that is currently reported may be a bit misleading.

hasher.go
```diff
@@ -113,42 +117,107 @@ func (n *Hasher) EmptyRoot() []byte {
...
+
+// validateSiblingsNamespaceOrder checks whether left and right as two sibling
+// nodes in an NMT have correct namespace IDs relative to each other, more
+// specifically, the maximum namespace ID of the left sibling should not exceed
+// the minimum namespace ID of the right sibling. Note that the function assumes
+// that the left and right nodes are in correct format, i.e., they are
+// namespaced hash values.
+func (n *Hasher) validateSiblingsNamespaceOrder(left, right []byte) (err error) {
+	// each NMT node has two namespace IDs for the min and max
+	totalNameSpaceLen := 2 * n.NamespaceLen
```
rootulp, line 178: [uber nit] in some scenarios the "S" in namespace is capitalized and in other scenarios it isn't. For codebase consistency, would it help to consolidate on no capitalization per https://en.wikipedia.org/wiki/Namespace

```suggestion
	totalNamespaceLen := 2 * n.NamespaceLen
```

rootulp (approved):
👍

hasher.go
staheri14: >  it seems possible to encounter this error when there are matching namespace ids but those namespace IDs are too short.

Thanks for the suggestion.  I primarily used `ErrMismatchedNamespaceSize` to be consistent with  the one used in here i.e., https://github.com/celestiaorg/nmt/blob/d40e66cb5d2edfeae5d44e9b95fa71ac6c34aae7/nmt.go#L498 that signifies a similar condition. 
However, I understand your point and I agree that we should make some changes.  Going to rename it, please see https://github.com/celestiaorg/nmt/pull/113/commits/b1f17fb73be6696485d8ddc9caf1898d6ad03da4.

hasher.go
staheri14: Resolved https://github.com/celestiaorg/nmt/pull/113/commits/b1f17fb73be6696485d8ddc9caf1898d6ad03da4.

hasher_test.go
```diff
@@ -232,3 +233,194 @@ func TestHashNode_ChildrenNamespaceRange(t *testing.T) {
...
+			"valid node",
+			2,
+			[]byte{0, 0},
+			[]byte{1, 1},
+			[]byte{1, 2, 3, 4},
+			false,
+			nil,
+		},
+		{ // mismatched namespace size
+			"invalid node: mismatching namespace sizes",
```
rootulp, line 294:
```suggestion
		{
			"invalid node: length",
```

hasher.go
```diff
@@ -14,6 +16,11 @@ const (
 
 var _ hash.Hash = (*Hasher)(nil)
 
+var (
+	ErrUnorderedSiblings = errors.New("NMT sibling nodes should be ordered lexicographically by namespace IDs")
+	ErrInvalidNodeLen    = errors.New("Invalid NMT node size")
```
rootulp, line 21: [nit][optional] 

https://github.com/golang/go/wiki/CodeReviewComments#error-strings

```suggestion
	ErrInvalidNodeLen    = errors.New("invalid NMT node size")
```

evan-forbes (approved):
> It is important to note that returning an error from this method is not possible as it would violate the interface it is intended to fulfill i.e., NodeHasher and Hash.

> The validation utility functions developed in this PR must be further integrated into both the tree construction and proof verification process to detect invalid inputs and notify the caller of any errors.

I'm wondering about the first statement. We might need to break the api, and that's ok imo. In fact, I assumed that we would have to to close #109. It would require significantly changes downstream, but I'm not sure that's avoidable even if we don't break the API. 

I guess I just have questions over how we actually plan to call these ahead of time in order to avoid panics, but we can discuss that synchronously if that would be faster :slightly_smiling_face:. Ideally, we would incorporate this into the code so that this was the default, since the panics here are purely to handle errors. We're still going to have to handle those errors downstream in order to write safe code, so might as well handle it here imo

otherwise LGTM, I like this change

hasher.go
staheri14: Addressed https://github.com/celestiaorg/nmt/pull/113/commits/faf4e5082ded0e955fad0bea46cef7e0c50b89a2

hasher_test.go
staheri14: Addressed https://github.com/celestiaorg/nmt/pull/113/commits/faf4e5082ded0e955fad0bea46cef7e0c50b89a2

staheri14:
> I guess I just have questions over how we actually plan to call these ahead of time in order to avoid panics, but we can discuss that synchronously if that would be faster 

I agree that some of the APIs will inevitably have to change. However, my aim was to keep the `HashNode` and `HashLeaf` signatures intact, instead, delegate error handling to the caller methods i.e., `Push`, and  proof verification. That is, they will check the nodes validity before calling HashNode and HashLeaf, and if there was an invalid node, they will return an error.
I can elaborate on this during our 1:1.

PR #134, opened 2023-03-14
Reworded explanation of calculating namespace when IgnoreMaxNamespace is set 

### LINKED ISSUE #133: Unclear explanation of namespace calculation when IgnoreMaxNamespace is set
The section which explains how to calculate the namespace range when the parameter`IgnoreMaxNamespace` can be confusing. ([here](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/spec/nmt.md#ignore-max-namespace)).
In particular, the part 

> This is achieved by taking the maximum value among the namespace IDs available in the range of node's left and right children (i.e., n.maxNs = max(l.minNs, l.maxNs , r.minNs, r.maxNs)), which is not equal to the maximum possible namespace ID value.

could be read as suggesting that the calculated `n.maxNs` is necessarily no equal to the maximum possible namespace ID value.

ivan-gavran:
## Overview

This PR rephrases the explanation on how the namespace range is calculated when the parameter `IgnoreMaxNamespace` is set. If merged, it will close #133 . 

I tried to simplify sentences explaining the logic and add a precise expression at the end of the explanation.

[rendered document](https://github.com/ivan-gavran/nmt/blob/ivan/ignoreMaxNsExplanation/docs/nmt-lib.md)

## Checklist

### DIFF
--- a/docs/nmt-lib.md
+++ b/docs/nmt-lib.md
@@ -44,11 +44,13 @@ idSize := tree.NamespaceSize() // outputs 1
 
 ### Ignore Max Namespace
 
-If the NMT is configured with `IgnoreMaxNamespace` set to true, then the calculation of the namespace ID range of non-leaf nodes in the [namespace hash function](./spec/nmt.md#namespaced-hash) will change slightly.
-That is, when determining the upper limit of the namespace ID range for a tree node `n`, with children `l` and `r`, the maximum possible namespace ID, equivalent to `NamespaceIDSize()` bytes of `0xFF`, or `2^NamespaceIDSize()-1`,
-should be omitted if feasible (in the preceding code example with the ID size of `1` byte, the maximum possible namespace ID would be `0xFF`).
-This is achieved by taking the maximum value among the namespace IDs available in the range of node's left and right children (i.e., `n.maxNs = max(l.minNs, l.maxNs , r.minNs, r.maxNs))`, which is not equal to the maximum possible namespace ID value.
-If such a namespace ID does not exist, the `maxNs` is calculated as normal, i.e., `n.maxNs = max(l.maxNs , r.maxNs)`.
+If the NMT is configured with `IgnoreMaxNamespace` set to true (the flag is explained [here](#nmt-initialization-and-configuration)), then the calculation of the namespace ID range of non-leaf nodes in the [namespace hash function](./spec/nmt.md#namespaced-hash) will change slightly.
+That is, when determining the upper limit of the namespace ID range for a tree node, the maximum possible namespace `maxPossibleNamespace` should not be taken into account. 
+(In the preceding code example with the ID size of `1` byte, the value of `maxPossibleNamespace` is $2^8-1 = 0xFF$.)
+
+Concretely, for a node `n` with children `l` and `r`, the namespace ID is the largest namespace value from `l` and `r` different from  `maxPossibleNamespace`, if such a namespace ID exists.
+Otherwise, if all candidate values are equal to `maxPossibleNamespace`, the namespace ID of `n` is set to `maxPossibleNamespace`.
+Precisely, if a set `C = {ns \in {l.minNs, l.maxNs, r.minNs, r.maxNs}: ns < maxPossibleNamespace}` is not empty, `n.maxNs = max(C)`. If `C` is empty, `n.maxNs = maxPossibleNamespace`.
 
 ## Add Leaves
 
--- a/docs/nmt-lib.md
+++ b/docs/nmt-lib.md
@@ -48,7 +48,7 @@ If the NMT is configured with `IgnoreMaxNamespace` set to true (the flag is expl
 That is, when determining the upper limit of the namespace ID range for a tree node, the maximum possible namespace `maxPossibleNamespace` should not be taken into account.
 (In the preceding code example with the ID size of `1` byte, the value of `maxPossibleNamespace` is $2^8-1 = 0xFF$.)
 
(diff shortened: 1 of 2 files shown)


rootulp (approved):
👏 seems clearer to me, thanks!

docs/nmt-lib.md
```diff
@@ -44,11 +44,13 @@ idSize := tree.NamespaceSize() // outputs 1
...
+If the NMT is configured with `IgnoreMaxNamespace` set to true (the flag is explained [here](#nmt-initialization-and-configuration)), then the calculation of the namespace ID range of non-leaf nodes in the [namespace hash function](./spec/nmt.md#namespaced-hash) will change slightly.
+That is, when determining the upper limit of the namespace ID range for a tree node, the maximum possible namespace `maxPossibleNamespace` should not be taken into account. 
+(In the preceding code example with the ID size of `1` byte, the value of `maxPossibleNamespace` is $2^8-1 = 0xFF$.)
+
+Concretely, for a node `n` with children `l` and `r`, the namespace ID is the largest namespace value from `l` and `r` different from  `maxPossibleNamespace`, if such a namespace ID exists.
+Otherwise, if all candidate values are equal to `maxPossibleNamespace`, the namespace ID of `n` is set to `maxPossibleNamespace`.
+Precisely, if a set `C = {ns \in {l.minNs, l.maxNs, r.minNs, r.maxNs}: ns < maxPossibleNamespace}` is not empty, `n.maxNs = max(C)`. If `C` is empty, `n.maxNs = maxPossibleNamespace`.
```
staheri14, line 53: The `\in` does not seem to be rendered properly in the output file. As a suggestion, it can be replaced with the word `in`.

staheri14 (approved):
Thanks, it looks clearer to me too! I added a small suggestion below.

docs/nmt-lib.md
```diff
@@ -44,11 +44,13 @@ idSize := tree.NamespaceSize() // outputs 1
...
-should be omitted if feasible (in the preceding code example with the ID size of `1` byte, the maximum possible namespace ID would be `0xFF`).
-This is achieved by taking the maximum value among the namespace IDs available in the range of node's left and right children (i.e., `n.maxNs = max(l.minNs, l.maxNs , r.minNs, r.maxNs))`, which is not equal to the maximum possible namespace ID value.
-If such a namespace ID does not exist, the `maxNs` is calculated as normal, i.e., `n.maxNs = max(l.maxNs , r.maxNs)`.
+If the NMT is configured with `IgnoreMaxNamespace` set to true (the flag is explained [here](#nmt-initialization-and-configuration)), then the calculation of the namespace ID range of non-leaf nodes in the [namespace hash function](./spec/nmt.md#namespaced-hash) will change slightly.
+That is, when determining the upper limit of the namespace ID range for a tree node, the maximum possible namespace `maxPossibleNamespace` should not be taken into account.
```
evan-forbes, line 48: quick linter fix
```suggestion
That is, when determining the upper limit of the namespace ID range for a tree node, the maximum possible namespace `maxPossibleNamespace` should not be taken into account.
```

docs/nmt-lib.md
ivan-gavran: Thanks @staheri14 ! 
I pushed a commit in which I used latex rendering only for \in. I am not sure that it looks good, though (there is a difference in rendering between that and other symbols, e.g., `a `$\in$` A`.

As a second option, I wrapped everything in latex (please check how the rendered doc [here](https://github.com/ivan-gavran/nmt/blob/ivan/ignoreMaxNsExplanationLatex/docs/nmt-lib.md)). This looks much better when rendered but introduces perhaps unnecessary formalism in writing.

Which of the two options you find better?  (There is also the third option, which you originally suggested, to simply use `in`, though I'm afraid that might be a bit confusing.)

docs/nmt-lib.md
```diff
@@ -44,11 +44,13 @@ idSize := tree.NamespaceSize() // outputs 1
...
+(In the preceding code example with the ID size of `1` byte, the value of `maxPossibleNamespace` is $2^8-1 = 0xFF$.)
+
+Concretely, for a node `n` with children `l` and `r`, the namespace ID is the largest namespace value from `l` and `r` smaller than  `maxPossibleNamespace`, if such a namespace ID exists.
+Otherwise, if all candidate values are equal to `maxPossibleNamespace`, the namespace ID of `n` is set to `maxPossibleNamespace`.
+Precisely, if a set `C = {ns `$\in$` {l.minNs, l.maxNs, r.minNs, r.maxNs}: ns < maxPossibleNamespace}` is not empty, `n.maxNs = max(C)`. If `C` is empty, `n.maxNs = maxPossibleNamespace`.
```
evan-forbes, line 53: we just need to fix the linter errors, and then we should be able to merge :slightly_smiling_face: 

```
docs/nmt-lib.md:53:36 MD038/no-space-in-code Spaces inside code span elements [Context: "` {l.minNs, l.maxNs, r.minNs, ..."]
docs/nmt-lib.md:53:21 MD038/no-space-in-code Spaces inside code span elements [Context: "`C = {ns `"]
```

docs/nmt-lib.md
staheri14: The LaTeX form of the math symbol `\in` is more readable. As an alternative approach, we could keep the math symbols in LaTeX format while leaving the variables as normal text.

```
`C` $= {$ `ns` $\in {$`l.minNs`, `l.maxNs`, `r.minNs`, `r.maxNs` $}:$ `ns` $<$ `maxPossibleNamespace` $}$
```

ivan-gavran: I like that - pushed a commit. (Also added different braces sizes.)

evan-forbes (approved):

PR #135, opened 2023-03-14
chore: upgrade to Go 1.19

rootulp:
Go 1.18 is no longer supported (see [endoflife](https://endoflife.date/go)) so this PR upgrades to Go 1.19. We could also upgrade directly to Go 1.20 but I opted for 1.19 because most other celestiaorg repos are on Go 1.19.


staheri14 (approved):
Thanks! 🚀

PR #136, opened 2023-03-14
feat!:  updates the HashNode and HashLeaf methods to return error instead of panic and refactors the code

### LINKED ISSUE #109: feat: Handling Panics in HashNode Method

staheri14:
## Overview
Closes #109 and #99.
This pull request ensures that the `HashNode` and `HashLeaf` functions return errors instead of panicking, and that caller functions properly handle any emitted errors. When generating proofs, any hashing errors are propagated with additional context. During verification, hashing errors are treated as unsuccessful proof verification.

Note that in this PR, the term `Irrecoverable errors` used to describe errors that any number of retries can correct the them. 

## Checklist

### DIFF
--- a/hasher.go
+++ b/hasher.go
@@ -82,17 +82,15 @@ func (n *Hasher) Write(data []byte) (int, error) {
 		rightChild := data[n.Size():]
 		if err := n.ValidateNodes(leftChild, rightChild); err != nil {
 			return 0, err
-		} else {
-			n.tp = NodePrefix
 		}
+		n.tp = NodePrefix
 	// leaf nodes contain the namespace length and a share
 	default:
 		// validate leaf format
 		if err := n.ValidateLeaf(data); err != nil {
 			return 0, err
-		} else {
-			n.tp = LeafPrefix
 		}
+		n.tp = LeafPrefix
 	}
 
 	n.data = data
--- a/hasher.go
+++ b/hasher.go
@@ -77,6 +77,7 @@ func (n *Hasher) Write(data []byte) (int, error) {
 	switch ln {
 	// inner nodes are made up of the nmt hashes of the left and right children
 	case n.Size() * 2:
+		// check the format of the data
 		leftChild := data[:n.Size()]
 		rightChild := data[n.Size():]
 		if err := n.ValidateNodes(leftChild, rightChild); err != nil {
--- a/hasher.go
+++ b/hasher.go
@@ -19,6 +19,7 @@ var _ hash.Hash = (*Hasher)(nil)
 var (
 	ErrUnorderedSiblings = errors.New("NMT sibling nodes should be ordered lexicographically by namespace IDs")
 	ErrInvalidNodeLen    = errors.New("invalid NMT node size")
+	ErrInvalidLeafLen    = errors.New("invalid NMT leaf size")
 )
 
 type Hasher struct {
@@ -66,7 +67,7 @@ func (n *Hasher) Size() int {
 // Requires data of fixed size to match leaf or inner NMT nodes. Only a single
 // write is allowed.
 // It panics if more than one single write is attempted.
-// It returns error if the data does not match an NMT non-leaf node or leaf node.
+// If the data does not match an NMT non-leaf node or leaf node, an error will be returned.
 func (n *Hasher) Write(data []byte) (int, error) {
 	if n.data != nil {
 		panic("only a single Write is allowed")
@@ -96,8 +97,7 @@ func (n *Hasher) Write(data []byte) (int, error) {
 
 // Sum computes the hash. Does not append the given suffix, violating the
 // interface.
-// It may panic if the data being hashed is invalid.
-// This should never happen if the error returned by the Write method is checked before attempting the Sum method.
+// It may panic if the data being hashed is invalid. This should never happen since the Write method refuses an invalid data and errors out.
 func (n *Hasher) Sum([]byte) []byte {
 	switch n.tp {
 	case LeafPrefix:
@@ -145,7 +145,7 @@ func (n *Hasher) ValidateLeaf(data []byte) (err error) {
 	nidSize := int(n.NamespaceSize())
 	lenData := len(data)
 	if lenData < nidSize {
-		return fmt.Errorf("%w: got: %v, want >= %v", ErrInvalidNodeLen, lenData, nidSize)
+		return fmt.Errorf("%w: got: %v, want >= %v", ErrInvalidLeafLen, lenData, nidSize)
 	}
 	return nil
 }
@@ -153,8 +153,7 @@ func (n *Hasher) ValidateLeaf(data []byte) (err error) {
 // HashLeaf computes namespace hash of the namespaced data item `ndata` as
 // ns(ndata) || ns(ndata) || hash(leafPrefix || ndata), where ns(ndata) is the
 // namespaceID inside the data item namely leaf[:n.NamespaceLen]). Note that for
-// leaves minNs = maxNs = ns(leaf) = leaf[:NamespaceLen]. HashLeaf can return error if
-// the input is not properly namespaced.
+// leaves minNs = maxNs = ns(leaf) = leaf[:NamespaceLen]. HashLeaf can return the ErrInvalidNodeLen error if the input is not namespaced.
 //
 //nolint:errcheck
 func (n *Hasher) HashLeaf(ndata []byte) ([]byte, error) {
@@ -183,7 +182,7 @@ func (n *Hasher) HashLeaf(ndata []byte) ([]byte, error) {
 }
 
 // ValidateNodeFormat checks whether the supplied node conforms to the
-// namespaced hash format.
+// namespaced hash format and returns an error if it does not.
 func (n *Hasher) ValidateNodeFormat(node []byte) (err error) {
 	totalNamespaceLen := 2 * n.NamespaceLen
 	nodeLen := len(node)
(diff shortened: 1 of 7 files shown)


staheri14:
This PR is ready for review, with the exception of the code coverage percentage, which is currently 0.14% below the threshold. The uncovered lines are in two parts of the proof.go file, where the `HashLeaf` function emits an error. However, these situations should not occur in practice, as the input is already validated in earlier lines via `ValidateLeaf`, making it impossible for any test scenario to hit those lines. Despite this, I have chosen not to silence the errors, as doing so could make the code vulnerable to unwanted bugs if someone updates the `HashLeaf` logic without updating the `ValidateLeaf` logic as well.

hasher_test.go
```diff
@@ -385,41 +425,196 @@ func TestHashNodeWithValidateNodes(t *testing.T) {
...
+	}{
+		{
+			"invalid leaf",
+			NewNmtHasher(sha256.New(), 2, false),
+			[]byte{0},
+			true,
+			ErrInvalidLeafLen,
+		},
+		{
+			"invalid node: left.max > right.imn",
```
rootulp, line 494:
```suggestion
			"invalid node: left.max > right.min",
```

nmt_test.go
```diff
@@ -845,3 +870,272 @@ func exampleTreeWithEightLeaves() *NamespacedMerkleTree {
...
+		tree    *NamespacedMerkleTree
+		wantErr bool
+		errType error
+	}{
+		{"corrupt leaf", treeWithCorruptLeaf, true, ErrInvalidLeafLen},
+		{"unordered leaves", treeWithUnorderedLeaves, true, ErrUnorderedSiblings},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			_, err := tt.tree.Root()
```
rootulp, line 1070: [question] should this invoke computeRoot? It currently seems identical to the above test:

```go
func Test_Root_Error(t *testing.T) {
    // ...

			_, err := tt.tree.Root()
```

proof.go
```diff
@@ -154,20 +161,20 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, data [][]byte,
...
 				return false
 			}
-			leafData := append(
-				gotLeafNid, gotLeaf[nIDLen:]...,
-			)
 			// hash the leaf data
-			gotLeafHashes = append(gotLeafHashes, hashLeafFunc(leafData))
+			leafHash, err := hashLeafFunc(gotLeaf)
+			if err != nil { // this can never happen due to the initial validation of the leaf at the beginning of the loop
+				return false
```
rootulp, line 175: [optional] if this can never happen, should it `panic`?

proof.go
```diff
@@ -222,64 +241,87 @@ func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID na
...
+	hashes := make([][]byte, len(leavesWithoutNamespace))
+	for i, d := range leavesWithoutNamespace {
+		// prepend the namespace to the leaf data
 		leafData := append(
 			append(make([]byte, 0, len(d)+len(nid)), nid...), d...,
 		)
-		hashes[i] = nth.HashLeaf(leafData)
+		res, err := nth.HashLeaf(leafData)
+		if err != nil {
+			return false // this never can happen since the leafData is guaranteed to be namespaced
```
rootulp, line 315: [question] given this should never happen, should it `panic`?

proof_test.go
```diff
@@ -188,3 +207,92 @@ func TestProof_MultipleLeaves(t *testing.T) {
...
+	hasher := nmt.treeHasher
+	root, err := nmt.Root()
+	require.NoError(t, err)
+
+	// create nmt proof for namespace ID 4
+	nID4 := namespace.ID{4, 4}
+	proof4, err := nmt.ProveNamespace(nID4)
+	require.NoError(t, err)
+	// proof4 is the inclusion proof for the leaf at index 3
+	leaf4WONamespace := nmt.leaves[3][nmt.NamespaceSize():] // the VerifyInclusion function expects the leaf without the namespace ID
```
rootulp, line 271: [question] what does the variable name here mean? It appears to truncate some leading bytes from the leaf so maybe leaf 4 without namespace? If so, proposal to rename 
```suggestion
	// proof4 is the inclusion proof for the leaf at index 3
	leaf4WithoutNamespace := nmt.leaves[3][nmt.NamespaceSize():] // the VerifyInclusion function expects the leaf without the namespace ID
```

[question] if VerifyInclusion expects the leaf without the namespace, why does this tampered test data fail?

proof.go
```diff
@@ -222,64 +241,87 @@ func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID na
...
 }
 
 // VerifyInclusion checks that the inclusion proof is valid by using leaf data
-// and the provided proof to regenerate and compare the root. Note that the leaf
-// data should not contain the prefixed namespace, unlike the tree.Push method,
+// and the provided proof to regenerate and compare the root. Note that the leavesWithoutNamespace data should not contain the prefixed namespace, unlike the tree.Push method,
 // which takes prefixed data. All leaves implicitly have the same namespace ID:
 // `nid`.
-func (proof Proof) VerifyInclusion(h hash.Hash, nid namespace.ID, leaves [][]byte, root []byte) bool {
+func (proof Proof) VerifyInclusion(h hash.Hash, nid namespace.ID, leavesWithoutNamespace [][]byte, root []byte) bool {
```
rootulp, line 305: 👍 I like this parameter rename

rootulp (approved):
No blocking feedback, great work!

hasher.go
```diff
@@ -86,16 +99,25 @@ func (n *Hasher) Write(data []byte) (int, error) {
...
 // interface.
+// It may panic if the data being hashed is invalid. This should never happen since the Write method refuses an invalid data and errors out.
 func (n *Hasher) Sum([]byte) []byte {
 	switch n.tp {
 	case LeafPrefix:
-		return n.HashLeaf(n.data)
+		res, err := n.HashLeaf(n.data)
+		if err != nil {
+			panic(err) // this should never happen since the data is already validated in the Write method
+		}
```
evan-forbes, line 109: [question]
could this still be hit when verifying a proof since we're not calling the write method ahead of time?

evan-forbes (commented):
only one question atm, but will continue to review

proof_test.go
staheri14: >[question] what does the variable name here mean? It appears to truncate some leading bytes from the leaf so maybe leaf 4 without namespace? If so, proposal to rename

In this variable name i.e., `leaf4WONamespace`, the `WO` part is the abbreviation of `without`, however,  it looks like unclear (since you asked), so I'm going to change it to the full version.

nmt_test.go
staheri14: Thanks for finding this, please see the corrected version in this commit https://github.com/celestiaorg/nmt/pull/136/commits/3fd5dab14ea4b8cb0f3c3046cc07b7af8079a141

proof_test.go
staheri14: >[question] what does the variable name here mean? It appears to truncate some leading bytes from the leaf so maybe leaf 4 without namespace? If so, proposal to rename

Please see the new variable name in https://github.com/celestiaorg/nmt/pull/136/commits/3f1b523e9d87d4254ced051aaf2f0e9f070f73e2

hasher.go
staheri14: If the `Write` method is not called prior to the `Sum`, then the code panics due to the absence of the `tp` field of the `Hasher`. More details follow:
For the `Sum` to get through, the `tp` field of the hasher should be set to one of the `LeafPrefix` or `NodePrefix` based on which the hasher determines whether the input data belongs to a leaf or a non-leaf node. Without making call to the `Write` method, the `Hasher.tp` field remains unset, hence in the mentioned switch statement, none of the `LeafPrefix` and `NodePrefix` cases are true, and the code goes straight to the `default` part which panics due to unset `tp` field.

staheri14: Another point is that, it is not possible to call `Sum` without making a call to the `Write` method, as it corresponds to doing nothing (the `Hasher` state is empty). Doing so always results in panic due to undefined state https://github.com/celestiaorg/nmt/blob/85ec3c9e7c593bf95f48c8cf7665097cc5f00e26/hasher.go#L121.

proof.go
staheri14: The reason it cannot happen is that the leaf is validated earlier in the code. However, if that section of the code changes in the future, we may encounter an error at this line. If the leaf is found to be invalid, we should return false instead of panicking.

staheri14: Even when you check the line where the leaf gets validated, you will see that we return false.

proof.go
staheri14: The same argument as in [this](https://github.com/celestiaorg/nmt/pull/136#discussion_r1145349472) also applies in here.

evan-forbes (approved):
LGTM!

staheri14:
benchstat results:
```
goos: darwin
goarch: arm64
pkg: github.com/celestiaorg/nmt
                          │ master.txt  │        nmt-err-handling.txt        │
                          │   sec/op    │   sec/op     vs base               │
ComputeRoot/64-leaves-10    38.30µ ± 2%   38.93µ ± 4%       ~ (p=0.089 n=10)
ComputeRoot/128-leaves-10   74.73µ ± 1%   75.29µ ± 1%  +0.75% (p=0.023 n=10)
ComputeRoot/256-leaves-10   152.1µ ± 3%   152.6µ ± 1%       ~ (p=0.579 n=10)
geomean                     75.79µ        76.48µ       +0.91%

```

Issue #139, opened 2023-03-17
Computing leaf hashes should be done after the range check


andrijamitrovic23:
[computeLeafHashesIfNecessary](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/nmt.go#L168) should be done after the [range check](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/nmt.go#L171). This calculation could be time/resource consuming but unnecessarily called if the range is invalid.

andrijamitrovic23:
Similar issue to [Hasher initialization and reset should be done after validity checks](https://github.com/celestiaorg/nmt/issues/131).

staheri14:
Thanks for opening the issue, your point is right about the computation overhead, however, in this case, the [computeLeafHashesIfNecessary](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/nmt.go#L168) should not be moved as it affects the length of the `leafHashes` which is used in earlier lines.

PR #151, opened 2023-03-22
chore: defer compute

rootulp:
Closes https://github.com/celestiaorg/nmt/issues/139

### DIFF
--- a/nmt.go
+++ b/nmt.go
@@ -165,12 +165,12 @@ func (n *NamespacedMerkleTree) Prove(index int) (Proof, error) {
 // the HashNode method in the Hasher.
 func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
 	isMaxNsIgnored := n.treeHasher.IsMaxNamespaceIDIgnored()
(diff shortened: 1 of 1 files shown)


nmt.go
```diff
@@ -165,12 +165,12 @@ func (n *NamespacedMerkleTree) Prove(index int) (Proof, error) {
 // the HashNode method in the Hasher.
 func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
 	isMaxNsIgnored := n.treeHasher.IsMaxNamespaceIDIgnored()
-	n.computeLeafHashesIfNecessary()
 	// TODO: store nodes and re-use the hashes instead recomputing parts of the
 	// tree here
 	if start < 0 || start >= end || end > len(n.leafHashes) {
 		return NewEmptyRangeProof(isMaxNsIgnored), ErrInvalidRange
 	}
+	n.computeLeafHashesIfNecessary()
```
staheri14, line 173: This computation should not be moved as it affects the length of the `leafHashes` which is used in earlier lines. The original version of the code is correct.

Issue #123, opened 2023-03-08
Comment about the return value for `ProveNamespace` function does not correspond the code


andrijamitrovic23:
[Comment](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/nmt.go#L182-L183) above `ProveNamespace` function stated that an error is returned when the requested nID is not in the tree's range. 

It appears that this is not aligned with the [comment and the code](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/nmt.go#L211-L215) in the `ProveNamespace` function for that specific case. These do not mention or return an error (nill is returned).

andrijamitrovic23:
Probably a new type of error should be added as it is done in the [ProveRange](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/nmt.go#L172) function for an invalid range.

rootulp:
@staheri14 do you know why ProveNamespace returns an EmptyRangeProof if the requested namespace is outside the range of the NMT? I would expect that an error is sufficient.

rootulp:
Multiple options to resolve this issue:
1. https://github.com/celestiaorg/nmt/pull/167 is less controversial
2. https://github.com/celestiaorg/nmt/pull/166 takes the approach recommended in this issue by creating a new error

staheri14:
> do you know why ProveNamespace returns an EmptyRangeProof if the requested namespace is outside the range of the NMT? I would expect that an error is sufficient because a user can determine that their requested namespace falls outside the bounds of the NMT by inspecting the NMT's min / max namespace. In other words, what value does the EmptyRangeProof add?

The godoc does not align with the implementation, and should be revised.
After conducting additional investigation, I think it may not be necessary to return an error when a query is made for a namespace ID that falls outside the range of the tree's name ID. Such a query is not an erroneous input and can be considered a reasonable request, for which an empty response would be appropriate. Therefore, I believe that revising the godoc would be sufficient.

PR #167, opened 2023-03-30
fix: ProveNamespace godoc

rootulp:
Closes https://github.com/celestiaorg/nmt/issues/123

### DIFF
--- a/nmt.go
+++ b/nmt.go
@@ -185,7 +185,9 @@ func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
 // ProveNamespace returns a range proof for the given NamespaceID.
 //
 // case 1) If the namespace nID is out of the range of the tree's min and max
-// namespace, returns an empty range proof.
+// namespace i.e., (nID < n.minNID) or (n.maxNID < nID) ProveNamespace returns an empty
+// Proof with empty nodes and the range (0,0) i.e., Proof.start = 0 and
+// Proof.end = 0 to indicate that this namespace is not contained in the tree.
 //
 // case 2) If the namespace nID is within the range of the tree's min and max
 // namespace i.e., n.minNID<= n.ID <=n.maxNID and the tree does not have any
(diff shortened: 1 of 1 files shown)


nmt.go
```diff
@@ -185,10 +185,7 @@ func (n *NamespacedMerkleTree) ProveRange(start, end int) (Proof, error) {
 // ProveNamespace returns a range proof for the given NamespaceID.
 //
 // case 1) If the namespace nID is out of the range of the tree's min and max
-// namespace i.e., (nID < n.minNID) or (n.maxNID < nID) ProveNamespace returns
-// an error and does not generate any range proof, instead it returns an empty
-// Proof with empty nodes and the range (0,0) i.e., Proof.start = 0 and
-// Proof.end = 0 to indicate that this namespace is not contained in the tree.
```
staheri14, line 191: I think we just need to remove where we say it returns error, the rest is fine. 

```
case 1) If the namespace nID is out of the range of the tree's min and max
// namespace i.e., (nID < n.minNID) or (n.maxNID < nID) ProveNamespace returns an empty
// Proof with empty nodes and the range (0,0) i.e., Proof.start = 0 and
// Proof.end = 0 to indicate that this namespace is not contained in the tree.
```

rootulp: Sure. IMO it seems excessively verbose to explicitly state the contents of an empty share proof in this comment. I think a more maintainable approach would be to reference "an empty share proof" in this comment. If desired the function that generates an empty share proof can describe why it returns Proof.start = 0 and Proof.end = 0.

rootulp: resolved by https://github.com/celestiaorg/nmt/pull/167/commits/8854743ead3c84817a1cdbf4d94fcb0c5ebc2abb

staheri14: > If desired the function that generates an empty share proof can elaborate on the contents of an empty share proof

Agree, though we need to have it ( "an empty share proof" ) defined somewhere.

staheri14 (approved):

Issue #117, opened 2023-03-01
What are the benefits of `IgnoreMaxNamespace`?


rootulp:
## Context

https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/nmt.go#L33-L40

## Candidates
It appears to reduce the range of an NMT root. Does the reduction in range of an NMT root benefit light clients? Since celestia light clients download the data availability header which contains all row roots, it knows the range of namespaces for each row so it can determine an upper bound for the range of row 2 by looking at the minimum namespace ID of row 3 (even if `IgnoreMaxNamespace=false`).

One other benefit: does `IgnoreMaxNamespace=true` reduce the size of an NMT proof by one node?

## Questions
Is it necessary to support this feature?

liamsi:
We also discussed of putting this feature into the wrapper as it is kinda confusing without the use-case/Celestia context. That said, it would make the wrapper probably even more complex and confusing too though.

rootulp:
@staheri14 mentioned that the `Proof`s returned by the NMT assume this feature is enabled so if we remove the `IgnoreMaxNamespace` feature, we need to confirm that proofs still work.

rootulp:
FWIW this feature appears to have been added in https://github.com/celestiaorg/nmt/commit/8a5e16765dfe09539738aa62a813f70c7dced901 which doesn't list a rationale or motivating issue.

rootulp:
Issues that become obsolete if we remove this feature:
1. https://github.com/celestiaorg/nmt/issues/148
2. https://github.com/celestiaorg/nmt/issues/159

rootulp:
No longer a candidate because celestia-node assumes the EDS row roots are constructed with IgnoreMaxNamespace=true so that it can efficiently find which row roots that correspond to a set of namespaced rows.

PR #168, opened 2023-03-30
feat!: remove ignore max namespace

rootulp:
Related https://github.com/celestiaorg/nmt/issues/117

Experimental branch to see what breaks downstream if we removed the IgnoreMaxNamespace option.


rootulp:
No longer a candidate because celestia-node assumes the EDS row roots are constructed with `IgnoreMaxNamespace=true` so that it can efficiently find which row roots correspond to a set of namespaced rows.

Issue #158, opened 2023-03-24
TestNamespacedMerkleTree_calculateAbsenceIndex_Panic isn’t complete


andrijamitrovic23:
[TestNamespacedMerkleTree_calculateAbsenceIndex_Panic](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/nmt_test.go#L581) obviously isn’t complete. It does not push the data from test cases to the nmt. It passes (the tested function panics as expected) because it tests an empty tree so [this panic](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/nmt.go#L362) will be called in both cases. So this test does not test the given cases. Push function should be called for the given data.

One test case should added that tests the panic if the tree is empty, as it is done unintentionally in this test.

PR #169, opened 2023-03-30
test: fix TestNamespacedMerkleTree_calculateAbsenceIndex_Panic

rootulp:
Closes https://github.com/celestiaorg/nmt/issues/158

### DIFF
--- a/nmt_test.go
+++ b/nmt_test.go
@@ -596,12 +596,17 @@ func TestNamespacedMerkleTree_calculateAbsenceIndex_Panic(t *testing.T) {
 		nID      namespace.ID
 		pushData []namespaceDataPair
 	}{
+		{"empty tree", []byte{0, 0}, []namespaceDataPair{}},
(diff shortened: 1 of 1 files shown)


nmt_test.go
```diff
@@ -596,12 +596,17 @@ func TestNamespacedMerkleTree_calculateAbsenceIndex_Panic(t *testing.T) {
 		nID      namespace.ID
 		pushData []namespaceDataPair
 	}{
+		{"empty tree", []byte{0, 0}, []namespaceDataPair{}},
 		{"((0,0) == nID < minNID == (0,1))", []byte{0, 0}, generateLeafData(nidLen, 1, 3, []byte{})},
 		{"((0,3) == nID > maxNID == (0,2))", []byte{0, 3}, generateLeafData(nidLen, 1, 3, []byte{})},
```
staheri14, line 601: We could also modify the test names of these two to be consistent with the first one e.g., "non-empty tree with 3 leaves: ((0,0) == nID < minNID == (0,1))"

staheri14 (approved):
Looks good to me, added a suggestion as well.

Issue #143, opened 2023-03-17
redundant hashing in the computeRoot 


staheri14:
# Problem
The [`computeRoot`](https://github.com/celestiaorg/nmt/blob/95759767a88cfcc3605d448d53294bfd1ce336c4/nmt.go#L469) function currently calculates the root from the leaves even if if the leaves are already hashed and the result is available in the `leafHashes` of `NamespacedMerkleTree`. This can result in redundant hashing and is inconsistent with the purpose of the `leafHashes` field, which is intended to act as a cache for hashed leaves and save computational time.

# Acceptance Criteria
The `computeRoot` function should be modified to use the `leafHashes` field if it is non-empty.

staheri14:
Closed by https://github.com/celestiaorg/nmt/pull/172

PR #172, opened 2023-03-31
feat:  compute leaf hashes in the Push method

### LINKED ISSUE #108: feat: compute leaf hashes immediately after data items are pushed
# Problem

When adding data items to the tree, their hash is not directly calculated. Instead, every time we need to access the leaf hashes, we have to compare the number of data items added so far (which can be found in the [`leaves`](url) field of [`NamespaceMerkleTree`](https://github.com/celestiaorg/nmt/blob/master/nmt.go#L85)) with the leaf hashes (which are stored in the [`leafHashes`](https://github.com/celestiaorg/nmt/blob/master/nmt.go#L98) field of the same struct). For example, see [the computeRoot() method](https://github.com/celestiaorg/nmt/blob/master/nmt.go#L427) and [computeLeafHashesIfNecessary()  function](https://github.com/celestiaorg/nmt/blob/fa98a59cbf582b3a0880ba0222a49e3bf5032f6d/nmt.go#L518).
This can make debugging and code maintenance more challenging.

# Acceptance criteria 
To simplify the process, it would be better to hash the data items immediately after they are pushed. We can address this issue by implementing this feature in the `Push` method and refactoring the other parts of the code as necessary.

staheri14:
## Overview
Closes #108  and https://github.com/celestiaorg/nmt/issues/143

## Checklist

### DIFF
--- a/nmt.go
+++ b/nmt.go
@@ -474,14 +474,6 @@ func (n *NamespacedMerkleTree) computeRoot(start, end int) ([]byte, error) {
 		n.visit(rootHash)
 		return rootHash, nil
 	case 1:
-		// leafHash, err := n.treeHasher.HashLeaf(n.leaves[start])
-		// if err != nil { // this should never happen since leaves are added through the Push method, during which leaves formats are validated to make sure they are hashable.
-		// 	return nil, fmt.Errorf("failed to hash leaf: %w", err)
-		// }
-		// if len(n.leafHashes) < len(n.leaves) {
-		// 	n.leafHashes = append(n.leafHashes, leafHash)
-		// }
-
 		leafHash := make([]byte, len(n.leafHashes[start]))
 		copy(leafHash, n.leafHashes[start])
 		n.visit(leafHash, n.leaves[start])
--- a/nmt.go
+++ b/nmt.go
@@ -579,24 +579,6 @@ func (n *NamespacedMerkleTree) updateMinMaxID(id namespace.ID) {
 	}
 }
 
-// computes the leaf hashes if not already done in a previous call of
-// NamespacedMerkleTree.Root()
-// Any errors return by this method is irrecoverable and indicate an illegal state of the tree (n).
-func (n *NamespacedMerkleTree) computeLeafHashesIfNecessary() error {
-	// check whether all the hash of all the existing leaves are available
-	if len(n.leafHashes) < len(n.leaves) {
-		n.leafHashes = make([][]byte, len(n.leaves))
-		for i, leaf := range n.leaves {
-			res, err := n.treeHasher.HashLeaf(leaf)
-			if err != nil { // should never happen since the validity of leaves is checked in the Push method
-				return err
-			}
-			n.leafHashes[i] = res
-		}
-	}
-	return nil
-}
-
 type leafRange struct {
 	// start and end denote the indices of a leaf in the tree. start ranges from
 	// 0 up to the total number of leaves minus 1 end ranges from 1 up to the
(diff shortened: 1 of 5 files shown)


nmt.go
```diff
@@ -418,8 +412,12 @@ func (n *NamespacedMerkleTree) Push(namespacedData namespace.PrefixedData) error
 		return err
 	}
 
+	// compute the leaf hash
+	res := n.treeHasher.MustHashLeaf(namespacedData)
```
evan-forbes, line 416: why not bubble the error here if there ever is one?

evan-forbes (commented):
only two questions, otherwise LGTM this makes sense.

iirc, we have some benchmarks in this repo, have we compared this change against the old yet?

rootulp (dismissed):
LGTM

nmt.go
staheri14: There are two reasons why the error is not being bubbled up:

- In line 410, the data validity is checked through the `validateAndExtractNamespace` function call. As a result, the `HashLeaf` function call cannot produce an error. To handle this scenario, the `MustHashLeaf` function is introduced, which assumes that the data passed to it has already been validated. This approach is similar to the current situation and avoids any unnecessary error handling.

- If we were to use `HashLeaf` and bubble up the error, we would need to introduce an if block in our code i.e., the `Push` method. However, this if block would not be covered in our tests, leading to a reduction in test coverage percentage. Therefore, I have opted not to bubble up the error in this situation.

Please let me know if this addresses your question.

staheri14:
Please find the benchmark comparison below. It's worth noting that the change introduced by this PR doesn't affect the [`ComputeRoot()` benchmark test](https://github.com/celestiaorg/nmt/blob/a20a72346da3ca2543c266a6d70fca1592f90b7f/nmt_test.go#L683) result. This is because the time measurement captures the total time taken to push `N` items and then compute the root. Before this change, the items were hashed inside the `Root()` method, whereas now they are hashed within the `Push()` method. As a result, the total number of hashes remains unchanged, and therefore the performance for this particular benchmark test does not improve.

```
goos: darwin
goarch: arm64
pkg: github.com/celestiaorg/nmt
                          │   old.txt   │              new.txt               │
                          │   sec/op    │   sec/op     vs base               │
ComputeRoot/64-leaves-10    36.28µ ± 1%   38.10µ ± 0%  +5.00% (p=0.000 n=10)
ComputeRoot/128-leaves-10   72.23µ ± 0%   75.29µ ± 0%  +4.24% (p=0.000 n=10)
ComputeRoot/256-leaves-10   146.4µ ± 1%   152.7µ ± 1%  +4.30% (p=0.000 n=10)
geomean                     72.66µ        75.94µ       +4.51%
```

nmt.go
evan-forbes: that makes sense, but documenting it in the code for posterity would be a good addition imo. we could also just ignore the error if its impossible (and we have comments), but in general I think its always best to bubble even if we don't strictly need to. Its one less thing the reader or future contributors have to think about.

I'm not blocking on it though! will defer to the author :slightly_smiling_face:

evan-forbes (approved):

nmt.go
staheri14: Fair point, replaced MustHashLeaf with HashLeaf and bubbled up the error, please see https://github.com/celestiaorg/nmt/pull/172/commits/d60367b11a2a123c4fd06b9772cc30f78793b666

evan-forbes: Yeah, 100% merge imo

evan-forbes (approved):

staheri14:
@rootulp Can you please review the recent changes 🙏🏼

rootulp (approved):

Issue #173, opened 2023-04-04
NMT mainnet readiness


staheri14:
This meta-issue is created with the objective of determining the priority of the NMT issues concerning the mainnet. It is important to note the current list provides a snapshot of the issues and is subject to change as further issues are detected.

### In progress
- https://github.com/celestiaorg/nmt/issues/108
- https://github.com/celestiaorg/nmt/issues/128
- https://github.com/celestiaorg/nmt/issues/143

### Error handling [Mainnet]
- https://github.com/celestiaorg/nmt/issues/137
- https://github.com/celestiaorg/nmt/issues/138
- https://github.com/celestiaorg/nmt/issues/140
- https://github.com/celestiaorg/nmt/issues/141
- https://github.com/celestiaorg/nmt/issues/142
- https://github.com/celestiaorg/nmt/issues/157
- https://github.com/celestiaorg/nmt/issues/164

### Code correctness [Mainnet]
- https://github.com/celestiaorg/nmt/issues/129
- https://github.com/celestiaorg/nmt/issues/121
- https://github.com/celestiaorg/nmt/issues/148
- https://github.com/celestiaorg/nmt/issues/153
- https://github.com/celestiaorg/nmt/issues/159
- https://github.com/celestiaorg/nmt/issues/181
- https://github.com/celestiaorg/nmt/issues/196
- https://github.com/celestiaorg/nmt/issues/110 and https://github.com/celestiaorg/nmt/issues/118 

### Code cleanup
- https://github.com/celestiaorg/nmt/issues/130
- https://github.com/celestiaorg/nmt/issues/132
- https://github.com/celestiaorg/nmt/issues/160

### Docs
- https://github.com/celestiaorg/nmt/issues/98

### New features [Good for Mainnet, not necessary though]
- https://github.com/celestiaorg/nmt/issues/97

- ~~https://github.com/celestiaorg/nmt/issues/117~~
- https://github.com/celestiaorg/nmt/issues/161

staheri14:
Omitting the following two issues from this epic, and with than, we can close the current epic:
- https://github.com/celestiaorg/nmt/issues/183: While this issue is currently under investigation, it has not been deemed critical for this epic. The primary objective of this epic was to prepare the nmt module for the latest release of the core/app, which has already been accomplished.
[...]

PR #175, opened 2023-04-11
feat!: checks the exact size of the node in HashNode

### LINKED ISSUE #153: ValidateNode does not check the exact size of the non-leaf nodes
# Problem
In an NMT, non-leaf nodes all have a particular size that is `2*namespace size + hash output size`. 
The `ValidateNode` should take into account this size and validate nodes' size accordingly. However, this is not the case in the current implementation. 

# Acceptance Criteria
Incorporate the exact size check to the `ValidateNode` method together with proper unit tests.

staheri14:
## Overview
Closes #153 (part of https://github.com/celestiaorg/nmt/issues/173)

## Checklist

### DIFF
--- a/hasher.go
+++ b/hasher.go
@@ -194,9 +194,9 @@ func (n *Hasher) MustHashLeaf(ndata []byte) []byte {
 }
 
 // ValidateNodeFormat checks whether the supplied node conforms to the
-// namespaced hash format and returns an error if it does not. Specifically, it returns ErrInvalidNodeLen if the length of the node is less than the 2*namespace length which indicates it does not match the namespaced hash format.
+// namespaced hash format and  returns ErrInvalidNodeLen if not.
 func (n *Hasher) ValidateNodeFormat(node []byte) (err error) {
-	expectedNodeLen := n.Size() // each NMT node has two namespace IDs for the min and max
+	expectedNodeLen := n.Size()
 	nodeLen := len(node)
 	if nodeLen < int(expectedNodeLen) {
 		return fmt.Errorf("%w: got: %v, want >= %v", ErrInvalidNodeLen, nodeLen, expectedNodeLen)
--- a/hasher.go
+++ b/hasher.go
@@ -194,12 +194,12 @@ func (n *Hasher) MustHashLeaf(ndata []byte) []byte {
 }
 
 // ValidateNodeFormat checks whether the supplied node conforms to the
-// namespaced hash format and  returns ErrInvalidNodeLen if not.
+// namespaced hash format and returns ErrInvalidNodeLen if not.
 func (n *Hasher) ValidateNodeFormat(node []byte) (err error) {
 	expectedNodeLen := n.Size()
 	nodeLen := len(node)
 	if nodeLen != expectedNodeLen {
-		return fmt.Errorf("%w: got: %v, want >= %v", ErrInvalidNodeLen, nodeLen, expectedNodeLen)
+		return fmt.Errorf("%w: got: %v, want %v", ErrInvalidNodeLen, nodeLen, expectedNodeLen)
 	}
 	return nil
 }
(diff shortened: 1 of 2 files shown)


hasher.go
```diff
@@ -194,12 +194,12 @@ func (n *Hasher) MustHashLeaf(ndata []byte) []byte {
...
 // ValidateNodeFormat checks whether the supplied node conforms to the
-// namespaced hash format and returns an error if it does not. Specifically, it returns ErrInvalidNodeLen if the length of the node is less than the 2*namespace length which indicates it does not match the namespaced hash format.
+// namespaced hash format and  returns ErrInvalidNodeLen if not.
 func (n *Hasher) ValidateNodeFormat(node []byte) (err error) {
-	totalNamespaceLen := 2 * n.NamespaceLen
+	expectedNodeLen := n.Size()
 	nodeLen := len(node)
-	if nodeLen < int(totalNamespaceLen) {
-		return fmt.Errorf("%w: got: %v, want >= %v", ErrInvalidNodeLen, nodeLen, totalNamespaceLen)
+	if nodeLen < int(expectedNodeLen) {
```
evan-forbes, line 201: why do we only want to check `<` here instead of `!=`?

evan-forbes (commented):
nice, LGTM, only had a single question

this change seems like it could be consensus breaking if some nodes use it and others don't, so we might want to rename this PR to `feat!...` to indicate such a change

staheri14:
> this change seems like it could be consensus breaking if some nodes use it and others don't, so we might want to rename this PR to feat!... to indicate such a change

Right, updated the PR description.

evan-forbes (approved):
:+1: :+1:

hasher.go
```diff
@@ -194,12 +194,12 @@ func (n *Hasher) MustHashLeaf(ndata []byte) []byte {
 }
 
 // ValidateNodeFormat checks whether the supplied node conforms to the
-// namespaced hash format and returns an error if it does not. Specifically, it returns ErrInvalidNodeLen if the length of the node is less than the 2*namespace length which indicates it does not match the namespaced hash format.
+// namespaced hash format and  returns ErrInvalidNodeLen if not.
```
rootulp, line 197:
```suggestion
// namespaced hash format and returns ErrInvalidNodeLen if not.
```

hasher.go
```diff
@@ -194,12 +194,12 @@ func (n *Hasher) MustHashLeaf(ndata []byte) []byte {
...
 func (n *Hasher) ValidateNodeFormat(node []byte) (err error) {
-	totalNamespaceLen := 2 * n.NamespaceLen
+	expectedNodeLen := n.Size()
 	nodeLen := len(node)
-	if nodeLen < int(totalNamespaceLen) {
-		return fmt.Errorf("%w: got: %v, want >= %v", ErrInvalidNodeLen, nodeLen, totalNamespaceLen)
+	if nodeLen != expectedNodeLen {
+		return fmt.Errorf("%w: got: %v, want >= %v", ErrInvalidNodeLen, nodeLen, expectedNodeLen)
```
rootulp, line 202: [blocking] this error message is misleading b/c the conditional has been updated

```suggestion
		return fmt.Errorf("%w: got %v, want %v", ErrInvalidNodeLen, nodeLen, expectedNodeLen)
```

hasher_test.go
```diff
@@ -277,6 +314,9 @@ func TestValidateSiblingsNamespaceOrder(t *testing.T) {
 }
 
 func TestValidateNodeFormat(t *testing.T) {
```
rootulp, line 316: [question] should a test case be added that is > the expected non-leaf node size?

rootulp (changes requested):

hasher.go
staheri14: Addressed https://github.com/celestiaorg/nmt/pull/175/commits/76b27b47eddfeddecd17532f544971458a979c33

hasher_test.go
staheri14: Sure, done in https://github.com/celestiaorg/nmt/pull/175/commits/f5427e5c71271735866d20f093c25a5689aef6b2

rootulp (approved):

Issue #161, opened 2023-03-29
The VerifyInclusion() function does not check proof completeness 


staheri14:
# Problem
This issue was identified while working on https://github.com/celestiaorg/nmt/issues/128
The [`VerifyInclusion`](https://github.com/celestiaorg/nmt/blob/master/proof.go#L305) method does not check the completeness of the NMT proofs. It mostly tries to mimic a normal Merkle inclusion proof without consideration of namespaces. However, based on how it is being used in Celestia-app, the completeness check seems necessary.

**Usage of  `VerifyInclusion`  in Celestia-app**: 
The [`VerifyInclusion`](https://github.com/celestiaorg/nmt/blob/master/proof.go#L305) is invoked by the [`func (sp ShareProof) VerifyProof() bool`](https://github.com/celestiaorg/celestia-core/blob/7e4ca8dfc9b1233e635033440c39ede271993a77/types/share_proof.go#L121) which aims to verify the inclusion proofs of a set of shares with identical namespace ID to their corresponding rows roots of the extended data square.  As the `verifyCompleteness` paramteter defaults to `false` in the  [`VerifyInclusion`](https://github.com/celestiaorg/nmt/blob/master/proof.go#L305), the completeness of the supplied shares in the `VerifyProof()` is ignored. 

This issue is created to clarify whether this is an intended behaviour or not? and to document the rational in the code and in the NMT specification.

liamsi:
I think the motivation for this was  that `VerifyInclusion` could also be used to verify a sub-set of a namespace (say a continuous same-namespace range, which is not the whole namespace; as the whole namespace is supposed to be verified via `VerifyNamespace` instead). I can see that this is not clear and can easily be misused. 

@evan-forbes @Wondertan @renaynay: Are there any other usage of VerifyInclusion? Specifically, does the namespace-subset inclusion verification even make sense at all in any use-case?

staheri14:
I see, thanks for the clarification, it makes sense now. :pray:

However, the implementation of `VerifyInclusion` may still cause confusion for those without additional context about the usage of nmt in Celstia.
[...]

Issue #157, opened 2023-03-24
ValidateInclusion function can panic if it is called with an invalid nid


andrijamitrovic23:
If the [ValidateInclusion](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/proof.go#L265) function is called with an `nID` which size is not equal to the namespace size of the `proof.nodes`, the function could panic.

If this happened, the the [nth](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/proof.go#L266) hasher will be constructed with a different namespace size than the one that the proof was calculated with. This could lead to a panic in [verifyLeafHashes](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/proof.go#L275) function within the [HashNode](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/proof.go#L254) at this [panic](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/hasher.go#L246). Panic ocures  because `validateSiblingsNamespaceOrder` could return an `ErrUnorderedSiblings` error because:

```go
	leftMaxNs := namespace.ID(left[n.NamespaceLen:totalNamespaceLen])
	rightMinNs := namespace.ID(right[:n.NamespaceLen])
```
one of the `leftMaxNs` or `rightMinNs` is wrongly sliced due to different namespace sizes at their construction and thus parsed so that the following check is true and error is returned:

```go
	if rightMinNs.Less(leftMaxNs) {
		return fmt.Errorf("%w: the maximum namespace of the left child %x is greater than the min namespace of the right child %x", ErrUnorderedSiblings, leftMaxNs, rightMinNs)
	}
```

Simple check like this [one](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/proof.go#L124-L129) in the `VerifyNamespace` should be enough.

I have created a test for this issue. The test has such input data that leads to the mentioned panic. Test is a part of the following PR: https://github.com/celestiaorg/nmt/pull/156

PR #176, opened 2023-04-11
feat: ensures consistent namespace hash format for all tree node parameters in VerifyInclusion function

staheri14:
## Overview
Closes https://github.com/celestiaorg/nmt/issues/157 
Closes https://github.com/celestiaorg/nmt/issues/161
## Checklist


staheri14:
This PR has reduced the code coverage because it includes early consistency checks at the start of the `Verifyinclusio` function. As a result, certain code sections that were previously producing errors will no longer do so. Consequently, the corresponding lines will not be executed, leading to a decrease in code coverage.

rootulp (approved):

proof.go
```diff
@@ -302,9 +302,22 @@ func (proof Proof) verifyLeafHashes(nth *Hasher, verifyCompleteness bool, nID na
...
+	// check that the root is valid w.r.t the NMT hasher
+	if err := nth.ValidateNodeFormat(root); err != nil {
+		return false
+	}
+	// check that all the proof.nodes are valid w.r.t the NMT hasher
+	for _, node := range proof.nodes {
+		if err := nth.ValidateNodeFormat(node); err != nil {
+			return false
+		}
+	}
```
evan-forbes, line 319: [non-blocking question]
similar to what we have discussed with needing to throw and error after the proof was verified if the namepsaces are not in order, do we need to do the same here?

evan-forbes (approved):
nice!

proof.go
staheri14: I believe the issue you are referring to is https://github.com/celestiaorg/nmt/issues/97. Currently, if namespaces are not in order, the verification result is false without providing any error. However, we can choose to return a proper error message to indicate the root cause of verification failure, if this information would be useful for the caller. However, my current understanding is that returning an error from proof verification methods such as `VerifyInclusion` may not address the issue in question, as the problem is related to verifying that shares of a block are not out of order. To achieve this, the full node must construct NMTs from the block data and identify out-of-order leaves during tree construction.

evan-forbes: > To achieve this, the full node must construct NMTs from the block data and identify out-of-order leaves during tree construction.

ahh I See. consensus full nodes having the ability to generate otherwise valid proofs makes sense, but that's not possible, so there's no point. 

do we need to enable non-consensus full nodes to have this ability if we want to protect against this in a trust minimized way then?

evan-forbes: >  Please let me know if this is a desired feature and I can track it in a separate GH issue.

I think we need more discussion around this and to walk through the entire flow. this convo is definitely not blocking merging. 

I'm not entirely sure that we have to worry about it, since we this is the equivalent of 2/3s signing over an invalid data root. Typically, full nodes would simply reject that block, which is what would happen now, in our case though, this could result in light clients being tricked in very specific scenarios. It seems likely something that we would want to fix to have trust minimized exclusion proofs. meaning that when a consensus or full node returns all the data in a namespace along with a prove of inclusion and completeness, then they can rely on the completeness without making an honest majority assumption or running a consensus node.

staheri14: > I think we need more discussion around this and to walk through the entire flow.

I will make a note to discuss this further in our next call.

evan-forbes:
morge?

Issue #140, opened 2023-03-17
verifyLeafHashes function will panic if called from VerifyInclusion function over an empty proof


andrijamitrovic23:
[`verifyLeafHashes`](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/proof.go#L177) function will panic  when called over an empty proof. Panic will happen when [`getSplitPoint(proof.end)`](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/proof.go#L248) is called because an empty proof has proof.end = 0 that will lead to panic within `getSplitPoint` function [here](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/nmt.go#L454-L456).

This can happen if the [VerifyInclusion](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/proof.go#L265) function is called over an empty proof. This function will not process the empty proof before calling `verifyLeafHashes`.

In contrast to when called from `VerifyInclusion`, `verifyLeafHashes` will not be called with an empty proof from the `VerifyNamespace` which handles empty proof before calling `verifyLeafHashes` ([here](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/proof.go#L132) and [here](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/proof.go#L168)). 

`VerifyInclusion` should process the empty proof before calling `verifyLeafHashes`, but we suggest an additional protection in `verifyLeafHashes`.

PR #184, opened 2023-04-26
feat!: handles empty proofs in validateLeafHash and VerifyInclusion

staheri14:
## Overview
Closes https://github.com/celestiaorg/nmt/issues/140
The changes made in this PR are considered to be breaking, as they modify the behavior of `VerifyInclusion`. In the prior version, the `VerifyInclusion` function would return `false` for the first test case named ["valid empty proof and leaves == empty."](https://github.com/celestiaorg/nmt/blob/712ed08a521b254b62a707a3f2d041d3f3a8d83f/proof_test.go#L442) of the `TestVerifyInclusion_EmptyProofs` suite. However, with the changes implemented, an empty proof for an empty set of leaves is now considered a valid proof. As a result, `VerifyInclusion` now returns `true`.
cc: @liamsi 

## Checklist

### DIFF
--- a/nmt.go
+++ b/nmt.go
@@ -246,6 +246,7 @@ func (n *NamespacedMerkleTree) ProveNamespace(nID namespace.ID) (Proof, error) {
 }
 
 // validateRange validates the range [start, end) against the size of the tree.
+// start is inclusive and end is non-inclusive.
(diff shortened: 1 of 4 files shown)


rootulp (approved):

evan-forbes (approved):
LGTM

PR #187, opened 2023-05-01
feat: verifies the namespace order for each tree node

### LINKED ISSUE #129: Check ordering of namespace among child nodes in HashNode function
This [testCase](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/hasher_test.go#L93-L101) submits children nodes with invalid namespace ordering to HashNode function. The left child node has minNs larger than the maxNs of the same node, and in addition to that leftMinNs is even larger then the rightMinNs. The test will pass because HashNode function panics (in [validateSiblingsNamespaceOrder](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/hasher.go#L177-L194)) only if the rightMinNs is less than leftMaxNs.

Evan if the before mentioned test case is totally invalid based on the fact that the Push function will guarantee that the namespace ids will be ordered, it raises two questions or two paths for this issue:

1.  Should this test case be removed?
2. Should [validateSiblingsNamespaceOrder](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/hasher.go#L177-L194) check the ordering of leftMinNs, leftMaxNs, rightMinNs and rightMaxNs among children nodes?

I would suggest that the 'validateSiblingsNamespaceOrder' checks the ordering of the children namespaces just to be more robust, but the decisions is on the development team.

staheri14:
## Overview
Closes #129.

## Checklist


hasher_test.go
```diff
@@ -735,3 +769,73 @@ func Test_MustHashLeaf_Panic(t *testing.T) {
...
+
+	for _, ts := range tt {
+		t.Run(ts.name, func(t *testing.T) {
+			maxResult := max(ts.ns, ts.ns2)
+			assert.Equal(t, ts.expected, maxResult)
+		})
+	}
+}
+
+func TestMin(t *testing.T) {
```
staheri14, line 808: Added these test as the codcov was complaining about the lack of coverage for some part of the `min` function.

hasher_test.go
```diff
@@ -735,3 +769,73 @@ func Test_MustHashLeaf_Panic(t *testing.T) {
 		})
 	}
 }
+
+func TestMax(t *testing.T) {
```
staheri14, line 773: Added these tests for consistency with the tests developed for the `min` counterpart.

rootulp (approved):

evan-forbes (approved):

PR #188, opened 2023-05-01
feat: ensures consistent use of the assumption that NMT nodes are ordered ascendingly

### LINKED ISSUE #121: Assumption that leaves are ordered is not used consistently
There is an assumption that leaves of NMTs are ordered by their namespace ids.

However, this assumption is not used.
For instance, [here](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/spec/nmt.md?plain=1#L38) the spec takes `min(l.minNs, r.minNs)`, where `l.minNs` would suffice. Similarly, [here](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/hasher.go#L255) in the code the same thing happens. On the other hand, the assumption is used in the function [CalculateAbsenceIndex](https://github.com/celestiaorg/nmt/blob/4276d172f18c87ebdd18da0cc4b758f0dd164118/nmt.go#L355-L355), when finding the appropriate leaf for the absence proof.

As this seems to be an important assumption, it would probably make the code and the spec clearer if it were consistently used.

staheri14:
## Overview
Closes #121 and #148.
Please refer to this [PR](https://github.com/celestiaorg/nmt/pull/193) to compare the new and old version of namsespacre range calculation for a parent node in the `HashNode()`.

## Checklist


hasher.go
```diff
@@ -274,14 +283,10 @@ func (n *Hasher) HashNode(left, right []byte) ([]byte, error) {
 	leftMinNs, leftMaxNs := left[:n.NamespaceLen], left[n.NamespaceLen:flagLen]
 	rightMinNs, rightMaxNs := right[:n.NamespaceLen], right[n.NamespaceLen:flagLen]
 
-	minNs := min(leftMinNs, rightMinNs)
-	var maxNs []byte
-	if n.ignoreMaxNs && n.precomputedMaxNs.Equal(leftMinNs) {
-		maxNs = n.precomputedMaxNs
```
staheri14, line 280: If leftMinNs represents the maximum possible namespace ID, then by definition, rightMax must also be the maximum possible namespace ID. Therefore, this check appears to be unnecessary.

hasher_test.go
```diff
@@ -839,3 +839,111 @@ func TestMin(t *testing.T) {
...
+			expectedMaxNs: []byte{0x01},
+		},
+		{
+			ignoreMaxNs:   true,
+			leftMinNs:     []byte{0x00},
+			leftMaxNs:     []byte{0x01},
+			rightMinNs:    []byte{0x02},
+			rightMaxNs:    precomputedMaxNs,
+			expectedMinNs: []byte{0x00},
+			expectedMaxNs: precomputedMaxNs,
```
rootulp, line 886: [non-blocking][question] this test case has me puzzled. Based on the feature `ignoreMaxNs` I would expect max ns to be `[]byte{0x02}`.

rootulp (approved):

hasher_test.go
staheri14: Good question, the `IgnoreMaxNamespace` ignores the `MaxNs` from the namespace range of the parent node when the right  subtree (the right child) of the parent is filled with the leaves all with `MaxNs`. In this testcase, the right subtree is not populated only by `MaxNs` leaves, so, the condition is not met hence the normal range calculation takes place. This is how it is done in the original code as well, please see https://github.com/celestiaorg/nmt/blob/fd00c52175c48bad64d03444689162fb9c6bee41/hasher.go#L280-L288.

evan-forbes (approved):

hasher_test.go
rootulp: Thanks for explaining. I think we can def merge this PR. 

No concrete feedback but I think the `ignoreMaxNs` feature is confusing b/c it only applies under certain circumstances.

staheri14: I agree with the confusion part. It takes time for one's mental model to be shaped around the semantics of IgnoreMaxNamespace, especially without any context regarding its use case in Celestia. Without such context, it may never make sense.

> it only applies under certain circumstances.
To provide further clarity, this condition applies only under two circumstances. Firstly, when the tree has more than one leaf, and secondly, when the right half of a (parent) node is fully occupied by values that all have the Maximum namespace. In essence, the first condition can be deduced from the second one, so there is effectively only one condition.

liamsi: I think we should (eventually) move the ignoreNS logic to celestia somehow as it is really difficult to grasp in the context of the NMT only.

liamsi (approved):

hasher_test.go
staheri14: > I think we should (eventually) move the ignoreNS logic to celestia somehow as it is really difficult to grasp in the context of the NMT only.

Agree with this, created the tracking issue as well https://github.com/celestiaorg/nmt/issues/195

PR #192, opened 2023-05-05
feat!: add an additional constraint of empty leafHash field to the empty proof definition

### LINKED ISSUE #181: overlap between empty proofs and absence proofs
## Problem
 In NMTs, an empty `proof` is defined as a proof that satisfies two conditions: (1) it contains an empty range i.e., `proof.start == proof.end`, and (2) its `nodes` field is empty. However, currently, we do not check whether `proof.leafHash == 0`. This can lead to situations where a proof can be treated as both an absence proof and an empty proof, depending on the order of operations (depends on which type is checked first). Specifically, a proof that satisfies conditions (1) and (2) but has a non-empty `leafHash` will pass both [`IsEmptyProof()`](https://github.com/celestiaorg/nmt/blob/34cf84ca83cd7596504a2e809083b82b5775bcfe/proof.go#L110) and [`IsOfAbsence()`](https://github.com/celestiaorg/nmt/blob/34cf84ca83cd7596504a2e809083b82b5775bcfe/proof.go#L68) checks.

The following test demonstrates this inconsistency:
```go
	// create an NMT with 8 sequentially namespaced leaves, numbered from 1 to 8.
	tree := exampleNMT(1, 1, 2, 3, 4, 5, 6, 7, 8)
	root, err := tree.Root()
	assert.NoError(t, err)
	dummyValue := createByteSlice(tree.treeHasher.Size(), 0x01)
	// this proof is neither a well-formed absence proof nor a well-formed empty proof
	proof := Proof{start: 1, end: 1, leafHash: dummyValue, nodes: [][]byte{}}
	assert.True(t, proof.IsOfAbsence())  // this check passes while proof is not a well-formed absence proof, this result is misleading
	assert.True(t, proof.IsEmptyProof()) // this check passes while proof is not a well-formed empty proof, this result is misleading

	// the proof is not a proper empty proof, yet it can be used as a valid empty proof
	isVerified := proof.VerifyNamespace(tree.treeHasher.baseHasher, []byte{10}, [][]byte{}, root)
	assert.True(t, isVerified)
```

staheri14:
## Overview
Closes #181
The updates made by this PR will result in a breaking change, as empty proofs with a non-empty `leafHash` field will no longer be considered valid. This means that their verification through `VerifyNamespace` will fail.
Example:
```go
nIDSize := 1
tree := exampleNMT(nIDSize, 1, 2, 3, 4)
root, err := tree.Root()
require.NoError(t, err)
hasher := 
proof := Proof{
		start:    0,
		end:      0,
		nodes:    nil,
		leafHash: tree.leafHashes[0],
	}

res := proof.VerifyNamespace(tree.treeHasher.baseHasher, []byte{0}, [][]byte{}, root)
require.True(res) // this is false with the changes in this PR, however, previously, it was true
```

## Checklist

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -66,7 +66,7 @@ func (proof Proof) Nodes() [][]byte {
 // IsOfAbsence returns true if this proof proves the absence of leaves of a
 // namespace in the tree.
 func (proof Proof) IsOfAbsence() bool {
-	return len(proof.leafHash) > 0 && !(proof.start < 0) && (proof.end == proof.start+1)
+	return len(proof.leafHash) > 0
 }
 
 // LeafHash returns nil if the namespace has leaves in the NMT. In case the
--- a/proof.go
+++ b/proof.go
@@ -66,7 +66,7 @@ func (proof Proof) Nodes() [][]byte {
 // IsOfAbsence returns true if this proof proves the absence of leaves of a
 // namespace in the tree.
 func (proof Proof) IsOfAbsence() bool {
-	return len(proof.leafHash) > 0
+	return len(proof.leafHash) > 0 && !(proof.start < 0) && (proof.end == proof.start+1)
 }
 
 // LeafHash returns nil if the namespace has leaves in the NMT. In case the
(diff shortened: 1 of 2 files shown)


evan-forbes (approved):
makes sense, LGTM!

Issue #159, opened 2023-03-27
The updateMinMaxID method disregards the `IgnoreMaxNamespace` semantic


staheri14:
# Problem

The utility method [updateMinMaxID](https://github.com/celestiaorg/nmt/blob/eb012d08857e799fcb0ac7af39d17caa7ece97e1/nmt.go#L572)  in nmt.go does not consider the `IgnoreMaxNamespace` semantic when updating the [`maxNID`](https://github.com/celestiaorg/nmt/blob/master/nmt.go#L107) field of the `NamespacedMerkleTree` after each leaf is pushed using [Push](https://github.com/celestiaorg/nmt/blob/master/nmt.go#L425). This leads to inconsistency between the name ID range stored by `NamespacedMerkleTree.minID` and `NamespacedMerkleTree.maxID` and the name ID range extracted from the NMT via its [MaxNamespace() method](https://github.com/celestiaorg/nmt/blob/eb012d08857e799fcb0ac7af39d17caa7ece97e1/nmt.go#L458), which looks at the NMT's root name ID range affected by the `ignoreMaxNamespace` logic. 
Note that the values `NamespacedMerkleTree.minID` and `NamespacedMerkleTree.maxID` are used to compare queried NIDs when creating namespace proofs, as shown in [this part of the code](https://github.com/celestiaorg/nmt/blob/eb012d08857e799fcb0ac7af39d17caa7ece97e1/nmt.go#L220). This means that  queries for the max namespace ID (i.e., parity share name ID), wont be responded by an empty proof. This is against the original purpose of  the `IgnoreMaxNamespace` flag.

Below is a sample test to demonstrate the inconsistency set out in this issue:

```go
nidSize := 1
data := [][]byte{
	append(namespace.ID{1}, []byte("leaf_0")...),
	append(namespace.ID{2}, []byte("leaf_1")...),
	append(namespace.ID{3}, []byte("leaf_2")...),
	append(namespace.ID{4}, []byte("leaf_3")...),
	append(namespace.ID{math.MaxUint8}, []byte("leaf_4")...),
	append(namespace.ID{math.MaxUint8}, []byte("leaf_5")...),
	append(namespace.ID{math.MaxUint8}, []byte("leaf_6")...),
	append(namespace.ID{math.MaxUint8}, []byte("leaf_7")...),
}

tree := New(sha256.New(), NamespaceIDSize(nidSize), IgnoreMaxNamespace(true))
for _, d := range data {
[...]

PR #194, opened 2023-05-09
feat!: accounts for IgnoreMaxNamespace flag in the ProveNamespace method

staheri14:
## Overview
Closes https://github.com/celestiaorg/nmt/issues/159
## Checklist

### DIFF
--- a/nmt_test.go
+++ b/nmt_test.go
@@ -818,13 +818,13 @@ func TestMinMaxNamespace(t *testing.T) {
 	testCases := []testCase{
 		{
 			name:    "example tree with four leaves",
-			tree:    exampleNMT(1, 0, 0, 1, 3),
+			tree:    exampleNMT(1, true, 0, 0, 1, 3),
 			wantMin: namespace.ID{0},
 			wantMax: namespace.ID{3},
 		},
 		{
 			name:    "example tree with eight leaves",
-			tree:    exampleNMT(2, 1, 2, 3, 4, 5, 6, 7, 8),
+			tree:    exampleNMT(2, true, 1, 2, 3, 4, 5, 6, 7, 8),
 			wantMin: namespace.ID{1, 1},
 			wantMax: namespace.ID{8, 8},
 		},
@@ -843,8 +843,8 @@ func TestMinMaxNamespace(t *testing.T) {
 }
 
 // exampleNMT creates a new NamespacedMerkleTree with the given namespace ID size and leaf namespace IDs. Each byte in the leavesNIDs parameter corresponds to one leaf's namespace ID. If nidSize is greater than 1, the function repeats each NID in leavesNIDs nidSize times before prepending it to the leaf data.
-func exampleNMT(nidSize int, leavesNIDs ...byte) *NamespacedMerkleTree {
-	tree := New(sha256.New(), NamespaceIDSize(nidSize))
+func exampleNMT(nidSize int, ignoreMaxNamespace bool, leavesNIDs ...byte) *NamespacedMerkleTree {
+	tree := New(sha256.New(), NamespaceIDSize(nidSize), IgnoreMaxNamespace(ignoreMaxNamespace))
 	for i, nid := range leavesNIDs {
 		namespace := bytes.Repeat([]byte{nid}, nidSize)
 		d := append(namespace, []byte(fmt.Sprintf("leaf_%d", i))...)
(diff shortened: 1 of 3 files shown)


nmt_test.go
```diff
@@ -1102,3 +1105,28 @@ func Test_MinMaxNamespace_Err(t *testing.T) {
...
+	nidSize := 1
+	MaxNS := byte(math.MaxUint8)
+	tests := []struct {
+		name         string
+		nIDList      []byte
+		isEmptyProof bool
+	}{
+		{"tree with no leaf", []byte{}, true},
+		{"tree with one leaf with MaxNS", []byte{MaxNS}, false},
+		{"tree with two leaves, the right leaf has MaxNS", []byte{1, MaxNS}, true},
```
rootulp, line 1120: [question] sorry I don't quite understand why `tree with one leaf with MaxNS` results in `isEmptyProof=false` and `tree with two leaves, the right leaf has MaxNS` results in `isEmptyProof=true`. Under what circumstances is `isEmptyProof=true` expected?

rootulp (commented):
[no blocking feedback] but can't approve b/c I don't understand the test cases

nmt_test.go
```diff
@@ -1102,3 +1105,28 @@ func Test_MinMaxNamespace_Err(t *testing.T) {
 		})
 	}
 }
+
+// TestProveNamespace checks the output of the ProveNamespace method when queried for the maximum namespace ID.
+func TestProveNamespace_MaxNamespace(t *testing.T) {
```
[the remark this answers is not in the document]
staheri14: Thanks, fixed https://github.com/celestiaorg/nmt/pull/194/commits/42eeaaa72d66b9b614c16afa2e3bf355cbaf218a.

nmt_test.go
staheri14: Sure, here is why this is the case:
The empty proof is only returned under two conditions: 1) when the tree is empty and thus covers no namespace, and 2) when the queried nID is not within the range of the namespace of the **root** node (emphasis on the root node, not the entire tree).

The calculation of the namespace range of the root node (and all the intermediate nodes) depends on the `IgnoreMaxNamespace` flag. When this flag is set, the namespace range of the right child/sub-tree is ignored if it only consists of leaves with the maximum namespace ID.

In the test case with one leaf with `MaxNs`, the root node has no left or right child, and the namespace range of the root node is the same as the only leaf, i.e., [`MaxNs`, `MaxNs`]. In this case, since the queried nID is also `MaxNs`, then `ProveNamespace` returns a normal inclusion proof, not an empty proof.

However, in the other test case, where the tree has two leaves then the root node has two children. The right child contains only leaves with `MaxNs`, so it meets the `IgnoreMaxNamespace` condition. As a result, the namespace range of the root node is calculated based only on its left child, which is [`1`, `1`]. Therefore, as the queried nID is `MaxNs`, an empty proof is returned because it is not within the range of the root node.

Hope this addresses your question. Please let me know if more explanation is needed.

rootulp: Extremely helpful, thanks!

evan-forbes (approved):
LGTM

staheri14:
Adding the following note for the future reference: The codecov has decreased since previously, there was a possibility of encountering an error in the [`buildRangeProof`](https://github.com/celestiaorg/nmt/blob/5ca3019ed42b548c76aa8fa37010e0fa6e945811/nmt.go#L251) function call from within the [`ProveNamespace`](https://github.com/celestiaorg/nmt/blob/5ca3019ed42b548c76aa8fa37010e0fa6e945811/nmt.go#L213) function. This error would occur if we deliberately corrupted the state of the tree and then attempted to create namespace proofs using the corrupted tree (tests are available in [`Test_ProveNamespace_Err`](https://github.com/celestiaorg/nmt/blob/42eeaaa72d66b9b614c16afa2e3bf355cbaf218a/nmt_test.go#L953) ). However, we now add an early call to the [`Root()`](https://github.com/celestiaorg/nmt/blob/5ca3019ed42b548c76aa8fa37010e0fa6e945811/nmt.go#L222) function at the beginning of `ProveNamespace`, which effectively catches any corruption in the tree state and returns immediately. As a result, when we reach the [`buildRangeProof`](https://github.com/celestiaorg/nmt/blob/5ca3019ed42b548c76aa8fa37010e0fa6e945811/nmt.go#L251) line, the tree state has already been validated and guaranteed to be valid, preventing the `buildRangeProof `error from occurring

Issue #200, opened 2023-05-25
Pretty print namespace.ID


walldiss:
In celestia-node we rely a lot on nmt namespace.ID type and want to log it in multiple places in human readable manner(hex representation). Current implementation of `String()` method on `namespace.ID` does direct convertion to string type, and makes it impossible to read.
I wonder if other users of `nmt` pkg would benefit from ability to pretty print namespace.ID. If so there could be 2 options:
  - changing existing `String()` method (would break compatibility)
  - add extra method `HexString`.

rootulp:
> add extra method `HexString`.

Sounds like a good idea to me!

PR #201, opened 2023-05-25
Add HexString method for namespace id

walldiss:
## Overview

Implements https://github.com/celestiaorg/nmt/issues/200

### DIFF
--- a/namespace/id.go
+++ b/namespace/id.go
@@ -1,6 +1,9 @@
 package namespace
 
-import "bytes"
+import (
+	"bytes"
+	"encoding/hex"
+)
 
 type ID []byte
 
@@ -27,4 +30,9 @@ func (nid ID) Size() IDSize {
 // String stringifies the nid.
 func (nid ID) String() string {
 	return string(nid)
+}
+
+// HexString returns hexadecimal encoding of nid.
+func (nid ID) HexString() string {
+	return hex.EncodeToString(nid)
 }
--- a/namespace/id_test.go
+++ b/namespace/id_test.go
@@ -1,8 +1,9 @@
 package namespace
 
 import (
-	"github.com/stretchr/testify/require"
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestID_HexString(t *testing.T) {
--- a/namespace/id_test.go
+++ b/namespace/id_test.go
@@ -1,0 +1,11 @@
+package namespace
+
+import (
+	"github.com/stretchr/testify/require"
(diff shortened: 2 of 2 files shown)


evan-forbes (commented):
oooh nice! 

this might actually be more useful in https://github.com/celestiaorg/celestia-app/blob/578c71cbfe033c1b597b903d22fda06f772db816/pkg/namespace/namespace.go#L8

since I think this namespace doesn't include the namespace verison, correct @rootulp? which could lead to devs using the incorrect hex

rootulp:
This namespace does include the namespace version so I think we can include this change here.

namespace/id.go
```diff
@@ -28,3 +31,8 @@ func (nid ID) Size() IDSize {
 func (nid ID) String() string {
 	return string(nid)
 }
+
+// HexString returns hexadecimal encoding of nid.
+func (nid ID) HexString() string {
+	return hex.EncodeToString(nid)
```
rootulp, line 37: [blocking] I think Codecov may block this PR until a unit test is added to cover this line

Screenshot 2023-05-25 at 10 29 41 AM

Otherwise LGTM

Wondertan:
Any reason not to change the default behavior of String to be always hex encoded? The current String is sorta useless while having default human-readable support for Stringer everywhere is not, AFAIU. The current String implementation can still be done with string(ID) if anyone needs raw nid in string format. 

@evan-forbes, @rootulp, thoughts?

rootulp:
My rationale for recommending a new `HexString()` method is to avoid breaking compatability

> changing existing String() method (would break compatibility)

If we can verify that no consumers of this library rely on the existing string representation then I think a human readable string method seems like a good idea.

rootulp (approved):

staheri14 (approved):
LGTM!

staheri14:
@walldiss Could you please revise the title of the PR to capitalize `HexString`? otherwise, it might be misleading, suggesting that the introduced method is private.

evan-forbes (approved):

Wondertan:
> If we can verify that no consumers of this library rely on the existing string representation then I think we can make the breaking change and modify the existing String() method

I verified the usage in the node and it's used only in places where human-readable string is expected.

evan-forbes:
I'm fine with mergin both, but I tend to agree that returning hex makes a lot of sense. if we need a string() for some reason, we can just call `string(id)`

rootulp:
Merging b/c @walldiss doesn't have permission to.

PR #203, opened 2023-05-26
test: ExampleNamespacedMerkleTree

rootulp:
Previously the function `ExampleNamespacedMerkleTree` existed in nmt_test.go. AFAIK Go only runs unit tests for functions with the name `Test*` so I wouldn't expect `ExampleNamespacedMerkleTree` to run by the Go test runner. But it is run and fails [here](https://github.com/celestiaorg/nmt/actions/runs/5092193098/jobs/9153228914?pr=202#step:7:551).

This PR converts the example into a test. Defer to @staheri14 if it should be deleted instead.

### DIFF
--- a/nmt_test.go
+++ b/nmt_test.go
@@ -52,62 +52,45 @@ func newNamespaceDataPairRaw(nidSize int, data []byte) namespaceDataPair {
 	}
 }
 
-func ExampleNamespacedMerkleTree() {
-	// the tree will use this namespace size
-	nidSize := 1
-	// the leaves that will be pushed
+func TestExampleNamespacedMerkleTree(t *testing.T) {
 	data := [][]byte{
 		append(namespace.ID{0}, []byte("leaf_0")...),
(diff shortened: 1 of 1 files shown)


staheri14 (approved):
Thanks for converting it to a test, I suggest keeping it. 

[question] What changes made the test pass? why it was failing?

rootulp:
> [question] What changes made the test pass? why it was failing?

The test started failing in https://github.com/celestiaorg/nmt/actions/runs/5092193098/jobs/9153228914?pr=202#step:7:551 because the text output for the `Example*`s is slightly different from the new output.

The test wanted this text output:

https://github.com/celestiaorg/nmt/blob/9efc9bff0ac35eed215de67c11f00e8ccf5a2137/nmt_test.go#L107-L110

and instead it got:

```
	// Output:
	// Min namespace: 00
	// Max namespace: 01
	// Successfully verified namespace: 3030
```

The changes in this PR make this test behave more like a typical test.

evan-forbes (approved):
LGTM

PR #211, opened 2023-06-25
feat: prepare for testing fraudulant blocks by allowing for custom hashers and the ability to add leaves without validation

evan-forbes:
## Overview

Currently, the hasher is not swappable, this is only problematic when trying to create fraudulent block test cases, since the standard hasher prevents such behavior.

## Checklist

### DIFF
--- a/hasher.go
+++ b/hasher.go
@@ -23,6 +23,17 @@ var (
 	ErrInvalidNodeNamespaceOrder = errors.New("invalid NMT node namespace order")
 )
 
+// NmtHasher describes the interface nmts use to hash leafs and nodes.
+type NmtHasher interface {
+	IsMaxNamespaceIDIgnored() bool
+	NamespaceSize() namespace.IDSize
+	HashLeaf(data []byte) ([]byte, error)
+	HashNode(leftChild, rightChild []byte) ([]byte, error)
+	EmptyRoot() []byte
+}
+
+var _ NmtHasher = &Hasher{}
+
 type Hasher struct {
 	baseHasher   hash.Hash
 	NamespaceLen namespace.IDSize
--- a/hasher.go
+++ b/hasher.go
@@ -14,7 +14,7 @@ const (
 	NodePrefix = 1
 )
 
-var _ hash.Hash = (*Hasher)(nil)
+var _ hash.Hash = (*NmtHasher)(nil)
 
 var (
 	ErrUnorderedSiblings         = errors.New("NMT sibling nodes should be ordered lexicographically by namespace IDs")
@@ -24,7 +24,7 @@ var (
 )
 
 // NmtHasher describes the interface nmts use to hash leafs and nodes.
-type NmtHasher interface {
+type Hasher interface {
 	IsMaxNamespaceIDIgnored() bool
 	NamespaceSize() namespace.IDSize
 	HashLeaf(data []byte) ([]byte, error)
@@ -32,9 +32,9 @@ type NmtHasher interface {
 	EmptyRoot() []byte
 }
 
-var _ NmtHasher = &Hasher{}
+var _ Hasher = &NmtHasher{}
 
-type Hasher struct {
+type NmtHasher struct { //nolint:revive
 	baseHasher   hash.Hash
 	NamespaceLen namespace.IDSize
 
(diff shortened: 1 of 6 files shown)


nmt.go
```diff
@@ -128,9 +129,18 @@ func New(h hash.Hash, setters ...Option) *NamespacedMerkleTree {
...
-	treeHasher := NewNmtHasher(h, opts.NamespaceIDSize, opts.IgnoreMaxNamespace)
+
+	// first create the default hasher using the updated options
+	hasher := NewNmtHasher(h, opts.NamespaceIDSize, opts.IgnoreMaxNamespace)
+	opts.Hasher = hasher
+
+	// set the options a second time to replace the hasher if needed
+	for _, setter := range setters {
+		setter(opts)
+	}
```
evan-forbes, line 140: I hate doing this a second time, but I'm not sure what else to do without sacrificing expected behavior from the user. 

we can't set hasher using the default namespace id size, and then the user also be able to change the default namespace ID size. So we have to set the options first, then create the default hasher, then optionally replace the default hasher with a user specified one

nmt_test.go
```diff
@@ -1138,3 +1138,19 @@ func TestEmptyRoot_NMT(t *testing.T) {
...
+		append(namespace.ID{1}, []byte("leaf_3")...),
+	}
+	nidSize := 1
+	tree := New(sha256.New(), NamespaceIDSize(nidSize))
+
+	for _, d := range data {
+		err := tree.ForceAddLeaf(d)
+		assert.NoError(t, err)
+	}
+}
```
evan-forbes, line 1156: Most things break without also using a custom hasher that also doesn't validate things here, so the only other thing we could test for is that it does break, which is already tested. The test coverage tracker is too basic to know that tho

hasher.go
```diff
@@ -23,7 +23,18 @@ var (
 	ErrInvalidNodeNamespaceOrder = errors.New("invalid NMT node namespace order")
 )
 
-type Hasher struct {
+// NmtHasher describes the interface nmts use to hash leafs and nodes.
```
rootulp, line 26: [nit] consistency between interface name and name referenced in comment. Since there is an NmtHasher type below, I wonder if this comment meant to be above line 37? Alternatively: 

```suggestion
// Hasher describes the interface NMT uses to hash leafs and nodes.
```

nmt.go
```diff
@@ -491,6 +501,27 @@ func (n *NamespacedMerkleTree) MaxNamespace() (namespace.ID, error) {
 	return MaxNamespace(r, n.NamespaceSize()), nil
 }
 
+// ForceAddLeaf adds a namespaced data to the tree without validating its
+// namespace ID. This method should only be used by tests that are attempting to
+// create out of order trees. The default hasher will fail for trees that are
+// out of order.
+func (n *NamespacedMerkleTree) ForceAddLeaf(leaf namespace.PrefixedData) error {
```
rootulp, line 508: > This method should only be used by tests

[question] is it possible to un-export this function and retain the tests that use it? It seems possible b/c `nmt_test.go` is in the same package



```suggestion
func (n *NamespacedMerkleTree) forceAddLeaf(leaf namespace.PrefixedData) error {
```

Asking so that we can enforce that users don't use this function

rootulp: ^ disregard. I assume this PR is meant to be used in conjunction with subsequent PRs that do use this exported function

rootulp (approved):

staheri14 (approved):
LGTM in general. I added an optional comment. 
I understand the point about introducing the `Hasher` as an interface type. While it does provide flexibility to NMT users, it also opens up the potential for errors and issues. However, it will require additional time to thoroughly identify the specific issues that users may encounter without proper implementation of the `Hasher` interface. We may want to look into this later. As we are currently using our own implementation that should be fine.

hasher.go
```diff
@@ -23,7 +23,18 @@ var (
...
+type Hasher interface {
+	IsMaxNamespaceIDIgnored() bool
+	NamespaceSize() namespace.IDSize
+	HashLeaf(data []byte) ([]byte, error)
+	HashNode(leftChild, rightChild []byte) ([]byte, error)
+	EmptyRoot() []byte
```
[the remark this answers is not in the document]
staheri14: One update on this suggested approach that is for a Byzantine-malicious node, the `AreSorted` function will always return true regardless, hence allowing unordered leaves to be inserted into the tree.

staheri14 (approved):

hasher.go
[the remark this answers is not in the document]
evan-forbes: that's a good idea, I was thinking about that too but wasn't sure if it belonged in the hasher interface. I was initially going to use a separate function, similar to NodeVisitor, but the current mechanism of of checking if the leaves are ordered properly requires access to `NamespacedMerkleTree` internals

nmt.go
evan-forbes: yeah we have to use this in app to create malicious commitments by hand

evan-forbes:
> However, it will require additional time to thoroughly identify the specific issues that users may encounter without proper implementation of the Hasher interface. We may want to look into this later. As we are currently using our own implementation that should be fine.

pretty much the only reason I can think of to do this is to purposefully mess things up lolol

left a note in the godoc for the hasher interface to say just that

rootulp (approved):

hasher.go
evan-forbes: sorry, forgot to mention that this was fixed in https://github.com/celestiaorg/nmt/pull/211/commits/2f7f25ec5e01ec8cb9d24509de1e32b510dda70d

hasher.go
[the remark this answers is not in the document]
staheri14: > I was thinking about that too but wasn't sure if it belonged in the hasher interface. 

The hasher (before the changes introduced in this PR) also has relevant methods for checking the ordering of the namespaces within each node, or across nodes, e.g., [`validateSiblingsNamespaceOrder`](https://github.com/celestiaorg/nmt/blob/master/hasher.go#L236). So, the logic pertaining to the leaves order seems not limited to the `NamespacedMerkleTree` internals.

PR #214, opened 2023-06-27
feat(proof.go): adding JSON marshal/unmarshal

distractedm1nd:
Adds json marshaling/unmarshaling to `nmt.Proof`.

As found elsewhere in the go stdlib, `UnmarshalJSON` must be defined on the pointer receiver although `MarshalJSON` is a value receiver. 

This is needed because the fields are private. Does this need to be the case? If we make them public, JSON marshaling will work by default.

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -2,6 +2,7 @@ package nmt
 
 import (
 	"bytes"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"hash"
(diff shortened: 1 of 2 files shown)


proof.go
```diff
@@ -47,6 +48,39 @@ type Proof struct {
 	isMaxNamespaceIDIgnored bool
 }
 
+type jsonProof struct {
+	Start                   int      `json:"start"`
+	End                     int      `json:"end"`
+	Nodes                   [][]byte `json:"nodes"`
+	LeafHash                []byte   `json:"leaf_hash"`
```
vgonkivs, line 55: what do you think about adding `omitempty` tag for these fields?

evan-forbes, line 55: perhaps stupid question, but why not add the json tags directly to the proof struct?

evan-forbes (approved):
only had a single non blocking question

proof.go
distractedm1nd: Unfortunately json tags require exported fields. Double checked and get the error:

```
Struct field 'isMaxNamespaceIDIgnored' has 'json' tag but is not exported
```

cmwaters (approved):
👍

proof.go
evan-forbes: makes sense!

PR #233, opened 2023-09-06
fix: replacing jsonProof with pb.Proof

distractedm1nd:
jsonProof is not necessary when we already have pb.Proof, which caused issues because of mismatching field names.

Related: https://github.com/celestiaorg/celestia-node/issues/2631
Specifically: https://github.com/celestiaorg/celestia-node/issues/2631#issuecomment-1708048715

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -49,36 +49,28 @@ type Proof struct {
 	isMaxNamespaceIDIgnored bool
 }
 
-type jsonProof struct {
-	Start                   int      `json:"start"`
-	End                     int      `json:"end"`
-	Nodes                   [][]byte `json:"nodes"`
-	LeafHash                []byte   `json:"leaf_hash"`
-	IsMaxNamespaceIDIgnored bool     `json:"is_max_namespace_id_ignored"`
-}
-
 func (proof Proof) MarshalJSON() ([]byte, error) {
-	jsonProofObj := jsonProof{
-		Start:                   proof.start,
-		End:                     proof.end,
-		Nodes:                   proof.nodes,
-		LeafHash:                proof.leafHash,
-		IsMaxNamespaceIDIgnored: proof.isMaxNamespaceIDIgnored,
+	pbProofObj := pb.Proof{
+		Start:                 int64(proof.start),
+		End:                   int64(proof.end),
+		Nodes:                 proof.nodes,
+		LeafHash:              proof.leafHash,
+		IsMaxNamespaceIgnored: proof.isMaxNamespaceIDIgnored,
 	}
-	return json.Marshal(jsonProofObj)
+	return json.Marshal(pbProofObj)
 }
 
 func (proof *Proof) UnmarshalJSON(data []byte) error {
-	var jsonProofObj jsonProof
-	err := json.Unmarshal(data, &jsonProofObj)
+	var pbProof pb.Proof
+	err := json.Unmarshal(data, &pbProof)
 	if err != nil {
 		return err
 	}
-	proof.start = jsonProofObj.Start
-	proof.end = jsonProofObj.End
-	proof.nodes = jsonProofObj.Nodes
-	proof.leafHash = jsonProofObj.LeafHash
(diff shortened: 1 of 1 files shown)


renaynay (approved):

Wondertan (commented):
At this point, we should consider making all the fields on the Proof public (It might be breaking).

It's feels wrong to reuse pb definition for json marshalling, even though it works.

distractedm1nd:
I still disagree with making fields public only for the point of serialization. These are not fields that should be touched outside of the given methods

proof.go
```diff
@@ -49,36 +49,28 @@ type Proof struct {
...
 
-type jsonProof struct {
-	Start                   int      `json:"start"`
-	End                     int      `json:"end"`
-	Nodes                   [][]byte `json:"nodes"`
-	LeafHash                []byte   `json:"leaf_hash"`
-	IsMaxNamespaceIDIgnored bool     `json:"is_max_namespace_id_ignored"`
-}
-
 func (proof Proof) MarshalJSON() ([]byte, error) {
```
rootulp, line 52: [not blocking][silly question] do we need to maintain the `type Proof` defined in this file? It seems very similar to the Protobuf generated type in `pb/proof.pb.go` with a few exceptions:
- Start/End are int64 in the protobuf type instead of int
- IsMaxNamespaceIgnored is exported in the protobuf type

If we could remove the `type Proof` defined in this file, would that aid maintainability? I assume then we could marshal / unmarshal to JSON with https://pkg.go.dev/google.golang.org/protobuf/encoding/protojson

It doesn't have to be done in this PR b/c this is a fix and I'm suggesting a refactor.

My primary concern with maintaining two copies of the Proof type is that if we add a new field to the `proof.go` type, the compiler won't tell us to also add it to the protobuf Proof type.

rootulp (approved):
[nit] I think this should be marked breaking (i.e. `fix!:` in PR title) because it impacts JSON encoding observed in downstream repos (i.e. celestia-node).

[question] Do we need a unit test in this PR that the field names in the proof JSON encoding match what is expected or does that test belong in celestia-node?

proof.go
distractedm1nd: We wouldnt even need that protojson, it would already marshal/unmarshal to JSON just fine alone. 

The issue with this is that we cannot define methods onto pb.Proof since it is autogenerated and in a different package. Also: see my above comment. So I would keep type Proof

Wondertan:
Actually true, encapsulation for Proofs is more important for security.

> These are not fields that should be touched outside of the given methods

Using autogenerated proto for json completely solves consistency between json and proto, as it also autogenerate json tags(TIL), so it's not that bad after all.

distractedm1nd:
Yeah but the main problem is not being able to define methods on it

Wondertan:
I meant keeping native go struct

oblique:
What is the status on this?

distractedm1nd:
@rootulp Can we get this merged?

proof.go
```diff
@@ -49,36 +49,28 @@ type Proof struct {
...
+	err := json.Unmarshal(data, &pbProof)
 	if err != nil {
 		return err
 	}
-	proof.start = jsonProofObj.Start
-	proof.end = jsonProofObj.End
-	proof.nodes = jsonProofObj.Nodes
-	proof.leafHash = jsonProofObj.LeafHash
-	proof.isMaxNamespaceIDIgnored = jsonProofObj.IsMaxNamespaceIDIgnored
+	proof.start = int(pbProof.Start)
```
staheri14, line 69: Is it intentional that we didn't use `int64` here?

staheri14 (approved):

proof.go
rootulp: Yes. The `Proof` type has a field for `start` with type int. See https://github.com/celestiaorg/nmt/pull/233/files#diff-8900a93a6d474bd8973beabf67b04383bde602d3bf747ac747eaaae531bcf889R23

PR #253, opened 2024-05-13
feat!: adds input range check to optimize VerifyLeafHashes and VerifyInclusion methods

staheri14:
This PR optimizes the `VerifyLeafHashes` and `VerifyInclusion` methods by ensuring that the size of the provided leaves or leaf hashes matches the proof range and, if not, making an early return in order to prevent unnecessary hashing operations that would otherwise occur.
It introduces a breaking change by altering the behaviour of `VerifyLeafHashes` and `VerifyInclusion`. Previously, verification would succeed even if the provided leaves or leaf hashes exceeded the proof range, as long as the proof was valid and the extra leaves were appended at the end. In the new implementation, the verification will return false or an error in such cases.

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -133,7 +133,7 @@ func NewAbsenceProof(proofStart, proofEnd int, proofNodes [][]byte, leafHash []b
 	return Proof{proofStart, proofEnd, proofNodes, leafHash, ignoreMaxNamespace}
 }
 
-// IsEmptyProof checks whether the proof corresponds to an empty proof as defined in NMT specifications https://github.com/celestiaorg/nmt/blob/master/docs/spec/nmt.md.
+// IsEmptyProof checks whether the proof corresponds to an empty proof as defined in NMT specifications https://github.com/celestiaorg/nmt/blob/main/docs/spec/nmt.md.
 func (proof Proof) IsEmptyProof() bool {
 	return proof.start == proof.end && len(proof.nodes) == 0 && len(proof.leafHash) == 0
 }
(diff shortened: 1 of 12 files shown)


proof.go
```diff
@@ -13,7 +13,10 @@ import (
 )
 
 // ErrFailedCompletenessCheck indicates that the verification of a namespace proof failed due to the lack of completeness property.
-var ErrFailedCompletenessCheck = errors.New("failed completeness check")
+var (
+	ErrFailedCompletenessCheck = errors.New("failed completeness check")
+	ErrWrongLeafHashesSize     = errors.New("wrong leafHashes size")
+)
```
rootulp, line 19: Sorry can't provide suggestion but I think the previous Go doc should be moved

```go
var (
	// ErrFailedCompletenessCheck indicates that the verification of a namespace proof failed due to the lack of completeness property.
	ErrFailedCompletenessCheck = errors.New("failed completeness check")
	ErrWrongLeafHashesSize     = errors.New("wrong leafHashes size")
)

```

rootulp (approved):
LGTM

staheri14:
[To the reviewers] The codecov [error](https://github.com/celestiaorg/nmt/actions/runs/9068204992/job/24914967372?pr=254#step:8:32) is not specific to this PR, I am investigating it.

proof.go
staheri14: Thanks fixed it https://github.com/celestiaorg/nmt/pull/253/commits/b94a414c3fdd9dad617c5ba893f3f9b979792f78

rootulp (dismissed):

walldiss (approved):

staheri14:
Waiting on #255 to merge

rootulp (approved):

evan-forbes (approved):

PR #257, opened 2024-05-24
refactor: popIfNonEmpty for  leaf hashes

walldiss:
Use `popIfNonEmpty` instead of manual advancement of slice.


proof.go
```diff
@@ -344,10 +344,8 @@ func (proof Proof) VerifyLeafHashes(nth *NmtHasher, verifyCompleteness bool, nID
 			// if the leaf index falls within the proof range, pop and return a
 			// leaf
 			if proof.Start() <= start && start < proof.End() {
-				leafHash := leafHashes[0]
 				// advance leafHashes
```
cristaloleg, line 347: Useless comment?

staheri14 (approved):

Issue #267, opened 2024-07-25
`TestVerifySubtreeRootInclusion_recursionBug`


rootulp:
## Context

I'm trying to bump celestia-node to use celestia-app v2.0.0. It turns out the [big test cases](https://github.com/celestiaorg/celestia-node/pull/3453/commits/198d02ed49ee9bdb35204644929a37c2192f88c7) for `TestProveCommitmentAllCombinations` identified an issue in NMT.

https://github.com/celestiaorg/nmt/pull/260

## Problem

Failing unit test to illustrate the issue. This results in a panic because `computeRoot` is called over and over with: `start=19`, `end=20`, `k=0`.

```go
// TestVerifySubtreeRootInclusion_recursionBug is motivated by a failing test
// case in celestia-node
func TestVerifySubtreeRootInclusion_recursionBug(t *testing.T) {
	namespaceIds := bytes.Repeat([]byte{1}, 64)
	tree := exampleNMT(1, true, namespaceIds...)
	root, err := tree.Root()
	require.NoError(t, err)

	nmthasher := tree.treeHasher
	hasher := nmthasher.(*NmtHasher)
	subtreeRoot, err := tree.ComputeSubtreeRoot(0, 4)
	require.NoError(t, err)
	subtreeRoots := [][]byte{subtreeRoot, subtreeRoot, subtreeRoot, subtreeRoot, subtreeRoot, subtreeRoot, subtreeRoot}
	subtreeWidth := 8

	proof, err := tree.ProveRange(19, 64)
	require.NoError(t, err)

	require.NotPanics(t, func() {
		// This hits:
		// runtime: goroutine stack exceeds 1000000000-byte limit
		// runtime: sp=0x14020160480 stack=[0x14020160000, 0x14040160000]
		// fatal error: stack overflow
		_, err = proof.VerifySubtreeRootInclusion(hasher, subtreeRoots, subtreeWidth, root)
		require.NoError(t, err)
	})
}
```

## Proposal

1. Fix the implementation in NMT so that it isn't possible to recurse indefinitely
2. Create a new NMT release
3. Bump to it in all downstream repos

PR #268, opened 2024-07-26
fix: check the correctness of the leaf ranges

rach-id:
## Overview

Closes: https://github.com/celestiaorg/nmt/issues/267

## Summary by CodeRabbit

- **New Features**
	- Strengthened error handling for subtree root verification and leaf range calculations to improve robustness against invalid inputs.

- **Bug Fixes**
	- Enhanced validation processes to prevent incorrect operations and ensure reliable range processing.

- **Tests**
	- Refined test cases to focus on relevant scenarios and added new tests for improved coverage of edge cases related to error handling.

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -640,7 +640,7 @@ func nextLeafRange(currentStart, currentEnd, subtreeWidth int) (LeafRange, error
 	}
 	rangeEnd := currentStart + currentRange
 	idealTreeSize := nextSubtreeSize(uint64(currentStart), uint64(rangeEnd))
-	if idealTreeSize != rangeEnd-currentStart {
+	if currentStart+idealTreeSize != rangeEnd-1 {
 		// this will happen if the calculated range does not correctly reference an inner node in the tree.
 		return LeafRange{}, fmt.Errorf("provided subtree width %d doesn't allow creating a valid leaf range [%d, %d)", subtreeWidth, currentStart, rangeEnd)
 	}
--- a/proof.go
+++ b/proof.go
@@ -640,7 +640,7 @@ func nextLeafRange(currentStart, currentEnd, subtreeWidth int) (LeafRange, error
 	}
 	rangeEnd := currentStart + currentRange
 	idealTreeSize := nextSubtreeSize(uint64(currentStart), uint64(rangeEnd))
-	if currentStart+idealTreeSize != rangeEnd-1 {
+	if currentStart+idealTreeSize != rangeEnd {
 		// this will happen if the calculated range does not correctly reference an inner node in the tree.
 		return LeafRange{}, fmt.Errorf("provided subtree width %d doesn't allow creating a valid leaf range [%d, %d)", subtreeWidth, currentStart, rangeEnd)
 	}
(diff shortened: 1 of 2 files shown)


proof.go
```diff
@@ -528,6 +528,10 @@ func (proof Proof) VerifySubtreeRootInclusion(nth *NmtHasher, subtreeRoots [][]b
 			return popIfNonEmpty(&subtreeRoots), nil
 		}
 
+		if end-start == 1 {
```
rootulp, line 531: [question] what is the intuition around end-start == 1? If end is non-inclusive and start is inclusive and I want to refer to a single node (i.e. node at index 3) then I assume start=3, end=4. In this case end-start does = 1.

rootulp (approved):
One question

proof.go
rach-id: I guess I need to add a comment there to explain it more. 
At that level, we reached a leaf, and we looked in the subtree roots whether we have it, and also looked in the side nodes if we have it, but we can't find any node. This means that since the begining, the provided ranges do not reference valid inner nodes

proof_test.go
```diff
@@ -1828,3 +1839,31 @@ func TestVerifySubtreeRootInclusion(t *testing.T) {
...
+
+	proof, err := tree.ProveRange(19, 64)
+	require.NoError(t, err)
+
+	require.NotPanics(t, func() {
+		// This hits:
+		// runtime: goroutine stack exceeds 1000000000-byte limit
+		// runtime: sp=0x14020160480 stack=[0x14020160000, 0x14040160000]
+		// fatal error: stack overflow
+		_, err = proof.VerifySubtreeRootInclusion(hasher, subtreeRoots, subtreeWidth, root)
```
cmwaters, line 1866: Huh? Is this supposed to overflow?

rootulp: it's not supposed to overflow. The comment for the panic was from before the fix, maybe it can be updated to:

```
		// This previously hit:
		// runtime: goroutine stack exceeds 1000000000-byte limit
		// runtime: sp=0x14020160480 stack=[0x14020160000, 0x14040160000]
		// fatal error: stack overflow
		_, err = proof.VerifySubtreeRootInclusion(hasher, subtreeRoots, subtreeWidth, root)
```

now the test verifies that it doesn't panic but causes an error instead

proof.go
rootulp: ok thanks!

proof_test.go
rach-id:
```suggestion
		// This previously hits:
		// runtime: goroutine stack exceeds 1000000000-byte limit
		// runtime: sp=0x14020160480 stack=[0x14020160000, 0x14040160000]
		// fatal error: stack overflow
		_, err = proof.VerifySubtreeRootInclusion(hasher, subtreeRoots, subtreeWidth, root)
```

rootulp (approved):

proof.go
cmwaters: It would be nice if a comment was added here to explain the logic

cmwaters (approved):

proof.go
rach-id: https://github.com/celestiaorg/nmt/pull/272

PR #272, opened 2024-08-02
docs: add comment to leaf case in verify subtree roots inclusion

rach-id:

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -528,6 +528,9 @@ func (proof Proof) VerifySubtreeRootInclusion(nth *NmtHasher, subtreeRoots [][]b
 			return popIfNonEmpty(&subtreeRoots), nil
 		}
 
+		// at this level, we reached a leaf, but we couldn't find any range corresponding
+		// to needed leaf [start, end). 
+		// this means that the initial provided [start, end) range was invalid.
 		if end-start == 1 {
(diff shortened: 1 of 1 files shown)


proof.go
```diff
@@ -528,6 +528,9 @@ func (proof Proof) VerifySubtreeRootInclusion(nth *NmtHasher, subtreeRoots [][]b
 			return popIfNonEmpty(&subtreeRoots), nil
 		}
 
+		// at this level, we reached a leaf, but we couldn't find any range corresponding
+		// to needed leaf [start, end).
+		// this means that the initial provided [start, end) range was invalid.
 		if end-start == 1 {
 			return nil, fmt.Errorf("the provided range [%d, %d) does not reference a valid inner node", proof.start, proof.end)
```
rootulp, line 535: Is this comment only applicable inside the if statement?

```suggestion
		if end-start == 1 {
			// At this level, we reached a leaf, but we couldn't find any range corresponding
			// to needed leaf [start, end).
			// This means that the initial provided [start, end) range was invalid.
			return nil, fmt.Errorf("the provided range [%d, %d) does not reference a valid inner node", proof.start, proof.end)
```

rootulp (dismissed):

cmwaters (approved):

rootulp (approved):

PR #291, opened 2025-03-18
feat: refactor VerifyLeafHashes, VerifyNamespace, and VerifyInclusion to expose computing root and validations outside of Proof

### LINKED ISSUE #290: feat: Expose Root Computation from VerifyLeafHashes
### Feature Request: Expose Root Computation from VerifyLeafHashes

**Description:**
Currently, the `VerifyLeafHashes` function in the NMT library performs two operations internally:
1. Computing the root from provided proof and data.
2. Comparing the computed root with an existing root.

To facilitate use cases where root computation alone is required (without immediate comparison), we propose breaking down `VerifyLeafHashes` into two distinct functions. The main goal is to expose the root computation logic separately.

**Proposed Change:**
- Extract root computation into a separate, publicly accessible function, e.g., `ComputeRoot(proof, data)`.
- Update `VerifyLeafHashes` to utilize this new function internally:

```go
func ComputeRoot(proof Proof, data []byte) ([]byte, error) {
    // Existing logic from VerifyInclusion to compute root
}

func VerifyLeafHashes(proof Proof, data []byte, root []byte) (bool, error) {
    computedRoot, err := ComputeRoot(proof, data)
    if err != nil {
        return false, err
    }
    return bytes.Equal(root, computedRoot), nil
}
```

**Benefits:**
- Allows consumers to recompute roots separately from verification, enabling more flexible use-cases and simplifying proofs.
- Minimal change, reusing approximately 99% of existing code.

**Impact:**
- Non-breaking change, backward-compatible.
- Minor refactor only; no new logic introduced.

Feedback and suggestions are welcome.

gupadhyaya:
Fixes #290 

## Summary by CodeRabbit

- **New Features**
  - Enhanced data integrity verification with a refined cryptographic proof process and improved error handling.
- **Refactor**
  - Streamlined the underlying verification flow to simplify the validation logic and bolster system reliability.
  - Isolated root computation logic into a dedicated method for improved modularity.

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -141,8 +141,7 @@ func (proof Proof) IsEmptyProof() bool {
 	return proof.start == proof.end && len(proof.nodes) == 0 && len(proof.leafHash) == 0
 }
 
-// IsValidEmptyRangeProof checks whether the proof is a valid empty range proof.
-func (proof Proof) IsValidEmptyRangeProof(nth *NmtHasher, nID namespace.ID, root []byte, leaves [][]byte, checkNS bool) bool {
+func (proof Proof) isValidEmptyRangeProof(nth *NmtHasher, nID namespace.ID, root []byte, leaves [][]byte, checkNS bool) bool {
 	nIDLen := nID.Size()
 
 	if !proof.IsEmptyProof() || len(leaves) != 0 {
@@ -163,27 +162,36 @@ func (proof Proof) IsValidEmptyRangeProof(nth *NmtHasher, nID namespace.ID, root
 	return nID.Less(rootMin) || rootMax.Less(nID) || bytes.Equal(root, nth.EmptyRoot())
 }
 
-// ComputeLeafHashes computes the hashes of the given leaves by prepending the namespace ID to each leaf and hashing it using the provided NMT hasher.
-func ComputeLeafHashes(nth *NmtHasher, nid namespace.ID, leaves [][]byte, checkOrAppend bool) ([][]byte, error) {
-	nIDLen := nid.Size()
+// ComputeAndValidateLeafHashes validates and hashes a list of leaves using the provided NMT hasher.
+func ComputeAndValidateLeafHashes(nth *NmtHasher, nid namespace.ID, leaves [][]byte) ([][]byte, error) {
 	hashes := make([][]byte, len(leaves))
 	for i, d := range leaves {
-		leafData := d
-		if checkOrAppend {
-			if nth.ValidateLeaf(leafData) != nil {
-				return nil, fmt.Errorf("leaf hash does not match the NMT hasher's hash format")
-			}
-			// check whether the namespace ID of the data matches the queried nID
-			if leadNid := namespace.ID(leafData[:nIDLen]); !leadNid.Equal(nid) {
-				// conflicting namespace IDs in data
-				return nil, fmt.Errorf("leaf hash %x does not belong to namespace %x", leafData, nid)
-			}
-		} else {
-			// prepend the namespace to the leaf data
-			leafData = append(
-				append(make([]byte, 0, len(d)+len(nid)), nid...), d...,
-			)
+		if nth.ValidateLeaf(d) != nil {
+			return nil, fmt.Errorf("leaf hash does not match the NMT hasher's hash format")
 		}
+		// check whether the namespace ID of the data matches the queried nID
+		if leadNid := namespace.ID(d[:nid.Size()]); !leadNid.Equal(nid) {
+			// conflicting namespace IDs in data
+			return nil, fmt.Errorf("leaf hash %x does not belong to namespace %x", d, nid)
+		}
+		res, err := nth.HashLeaf(d)
+		if err != nil {
+			return nil, err
+		}
+		hashes[i] = res
+	}
+	return hashes, nil
+}
+
+// ComputePrefixedLeafHashes computes NMT leaf hashes for raw leaf data by prepending the given namespace ID.
+func ComputePrefixedLeafHashes(nth *NmtHasher, nid namespace.ID, leaves [][]byte) ([][]byte, error) {
+	hashes := make([][]byte, len(leaves))
+	for i, d := range leaves {
+		// prepend the namespace to the leaf data
+		leafData := append(
+			append(make([]byte, 0, len(d)+len(nid)), nid...), d...,
+		)
+
 		res, err := nth.HashLeaf(leafData)
 		if err != nil {
 			return nil, err
@@ -218,7 +226,7 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, leaves [][]byt
 
 	// if empty range proof, check that the proof is valid
 	if proof.start == proof.end {
-		return proof.IsValidEmptyRangeProof(nth, nID, root, leaves, true)
+		return proof.isValidEmptyRangeProof(nth, nID, root, leaves, true)
 	}
 
 	gotLeafHashes := make([][]byte, 0, len(leaves))
@@ -226,7 +234,7 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, leaves [][]byt
 		gotLeafHashes = append(gotLeafHashes, proof.leafHash)
 	} else {
 		var err error
-		gotLeafHashes, err = ComputeLeafHashes(nth, nID, leaves, true)
+		gotLeafHashes, err = ComputeAndValidateLeafHashes(nth, nID, leaves)
 		if err != nil {
 			return false
 		}
@@ -462,11 +470,11 @@ func (proof Proof) VerifyInclusion(h hash.Hash, nid namespace.ID, leavesWithoutN
 
 	// validate empty proof range
 	if proof.start == proof.end {
-		return proof.IsValidEmptyRangeProof(nth, nid, root, leavesWithoutNamespace, false)
+		return proof.isValidEmptyRangeProof(nth, nid, root, leavesWithoutNamespace, false)
 	}
 
 	// add namespace to all the leaves
-	hashes, err := ComputeLeafHashes(nth, nid, leavesWithoutNamespace, false)
+	hashes, err := ComputePrefixedLeafHashes(nth, nid, leavesWithoutNamespace)
 	if err != nil {
 		return false
 	}
--- a/proof.go
+++ b/proof.go
@@ -165,20 +165,20 @@ func (proof Proof) isValidEmptyRangeProof(nth *NmtHasher, nID namespace.ID, root
 // ComputeAndValidateLeafHashes validates and hashes a list of leaves using the provided NMT hasher.
 func ComputeAndValidateLeafHashes(nth *NmtHasher, nid namespace.ID, leaves [][]byte) ([][]byte, error) {
 	hashes := make([][]byte, len(leaves))
-	for i, d := range leaves {
-		if nth.ValidateLeaf(d) != nil {
-			return nil, fmt.Errorf("leaf hash does not match the NMT hasher's hash format")
+	for i, leaf := range leaves {
+		if nth.ValidateLeaf(leaf) != nil {
+			return nil, fmt.Errorf("invalid leaf data: does not contain the expected namespace prefix")
 		}
 		// check whether the namespace ID of the data matches the queried nID
-		if leadNid := namespace.ID(d[:nid.Size()]); !leadNid.Equal(nid) {
+		if leafNid := namespace.ID(leaf[:nid.Size()]); !leafNid.Equal(nid) {
 			// conflicting namespace IDs in data
-			return nil, fmt.Errorf("leaf hash %x does not belong to namespace %x", d, nid)
+			return nil, fmt.Errorf("leaf with namespace ID %x does not belong to expected namespace %x", leafNid, nid)
 		}
-		res, err := nth.HashLeaf(d)
+		hash, err := nth.HashLeaf(leaf)
 		if err != nil {
-			return nil, err
+			return nil, fmt.Errorf("failed to hash leaf: %w", err)
 		}
-		hashes[i] = res
+		hashes[i] = hash
 	}
 	return hashes, nil
 }
@@ -186,13 +186,13 @@ func ComputeAndValidateLeafHashes(nth *NmtHasher, nid namespace.ID, leaves [][]b
 // ComputePrefixedLeafHashes computes NMT leaf hashes for raw leaf data by prepending the given namespace ID.
 func ComputePrefixedLeafHashes(nth *NmtHasher, nid namespace.ID, leaves [][]byte) ([][]byte, error) {
 	hashes := make([][]byte, len(leaves))
-	for i, d := range leaves {
+	for i, leaf := range leaves {
 		// prepend the namespace to the leaf data and hash it
-		res, err := nth.HashLeaf(slices.Concat(nid, d))
+		hash, err := nth.HashLeaf(slices.Concat(nid, leaf))
 		if err != nil {
 			return nil, err
 		}
-		hashes[i] = res
+		hashes[i] = hash
 	}
 	return hashes, nil
 }
--- a/proof.go
+++ b/proof.go
@@ -338,6 +338,48 @@ func (proof Proof) ValidateCompleteness(nth *NmtHasher, nID namespace.ID) error 
 	return nil
 }
 
+// ComputeRootWithBasicValidation computes the Merkle root from a given proof and a set of leaf hashes,
+// performing basic validation steps prior to computing the root.
+//
+// If isNamespace is true, it additionally validates:
+//   - That the leaf hashes belong to the specified namespace.
+//   - That the proof is complete for the given namespace.
+//
+// Parameters:
+//   - nth: The NMT hasher instance used for validation and root computation.
+//   - nID: The namespace ID that the proof is expected to correspond to.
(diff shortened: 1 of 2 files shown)


proof.go
```diff
@@ -141,6 +141,58 @@ func (proof Proof) IsEmptyProof() bool {
 	return proof.start == proof.end && len(proof.nodes) == 0 && len(proof.leafHash) == 0
 }
 
+// IsValidEmptyRangeProof checks whether the proof is a valid empty range proof.
+func (proof Proof) IsValidEmptyRangeProof(nth *NmtHasher, nID namespace.ID, root []byte, leaves [][]byte, checkNS bool) bool {
```
vgonkivs, line 145: why do we need this method and how empty proof can be valid or not? I see you are using it under if `proof.start == proof.end ` condition, but proof.start can't be equal to proof.end conceptually, because `end` is a non-inclusive index of the leave.

vgonkivs: I don't think we should check the validity of an empty proof at all.

proof.go
```diff
@@ -141,6 +141,58 @@ func (proof Proof) IsEmptyProof() bool {
...
+
+	// empty proofs are always rejected unless 1) nID is outside the range of
+	// namespaces covered by the root 2) the root represents an empty tree, since
+	// it purports to cover the zero namespace but does not actually include
+	// any such nodes
+	return nID.Less(rootMin) || rootMax.Less(nID) || bytes.Equal(root, nth.EmptyRoot())
+}
+
+// ComputeLeafHashes computes the hashes of the given leaves by prepending the namespace ID to each leaf and hashing it using the provided NMT hasher.
+func ComputeLeafHashes(nth *NmtHasher, nid namespace.ID, leaves [][]byte, checkOrAppend bool) ([][]byte, error) {
```
vgonkivs, line 167: Having such` checkOrAppend bool` flags/modes makes the code harder to read and maintain in the future. The user of this method will have to dive into this code and understand what is it for. Also, I would verify leaves every time during computation.

proof.go
```diff
@@ -164,82 +216,22 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, leaves [][]byt
...
-			}
-			if bytes.Equal(root, nth.EmptyRoot()) {
-				return true
-			}
-			return false
-		}
-		// the proof range is empty, and invalid
-		return false
+	// if empty range proof, check that the proof is valid
+	if proof.start == proof.end {
```
vgonkivs, line 220: invalid condition. I guess we can assume as invalid in case proof.start == proof.end.

proof.go
```diff
@@ -248,64 +240,70 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, leaves [][]byt
...
-			minNsID := MinNamespace(leafHash, nth.NamespaceSize())
-			maxNsID := MaxNamespace(leafHash, nth.NamespaceSize())
-			if !nID.Equal(minNsID) || !nID.Equal(maxNsID) {
-				return false, fmt.Errorf("leaf hash %x does not belong to namespace %x", leafHash, nID)
-			}
+	return nil
+}
+
+// ValidateSingleNamespace ensures all leaf hashes belong to the expected namespace.
+func (proof Proof) ValidateSingleNamespace(nth *NmtHasher, nID namespace.ID, leafHashes [][]byte) error {
```
vgonkivs, line 294: just `ValidateNamespace`.

proof.go
```diff
@@ -248,64 +240,70 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, leaves [][]byt
...
+		minNsID := MinNamespace(leafHash, nth.NamespaceSize())
+		maxNsID := MaxNamespace(leafHash, nth.NamespaceSize())
+		if !nID.Equal(minNsID) || !nID.Equal(maxNsID) {
+			return fmt.Errorf("leaf hash %x does not belong to namespace %x", leafHash, nID)
 		}
 	}
+	return nil
+}
 
+// ValidateCompleteness checks that the proof is complete, i.e., it verifies
```
vgonkivs, line 305: `it verifies...`

proof.go
walldiss: I think it was in previous implementation so ratianal is backwards compatability to previous behavior. This is meant as non-breaking PR, but we might want to consider to break it in following PR. 

@gupadhyaya  Perhaps we should make this func as unexported for now?

proof.go
walldiss: Yeah seems like two different semantics. Input is also different: one takes in leaf with prepended ns and another one does not. Maybe just make two different funcs or even extract validatation/prepending from hashing?
Also similar to above might be worth to keep private func for now if it is not meant to be used outside of pkg

proof.go
gupadhyaya: will have to come back to this. for now, made this function private and retained the logic that was there previously. don't want to block the getrange PR on this. @vgonkivs

proof.go
gupadhyaya: split into two separate functions.

proof.go
gupadhyaya: as mentioned in the previous comment, have to revisit this.

proof.go
```diff
@@ -141,6 +141,66 @@ func (proof Proof) IsEmptyProof() bool {
 	return proof.start == proof.end && len(proof.nodes) == 0 && len(proof.leafHash) == 0
 }
 
+func (proof Proof) isValidEmptyRangeProof(nth *NmtHasher, nID namespace.ID, root []byte, leaves [][]byte, checkNS bool) bool {
+	nIDLen := nID.Size()
```
cristaloleg, line 145: Nit: can be created after the 2nd `if`.

proof.go
```diff
@@ -141,6 +141,66 @@ func (proof Proof) IsEmptyProof() bool {
...
+	return hashes, nil
+}
+
+// ComputePrefixedLeafHashes computes NMT leaf hashes for raw leaf data by prepending the given namespace ID.
+func ComputePrefixedLeafHashes(nth *NmtHasher, nid namespace.ID, leaves [][]byte) ([][]byte, error) {
+	hashes := make([][]byte, len(leaves))
+	for i, d := range leaves {
+		// prepend the namespace to the leaf data
+		leafData := append(
+			append(make([]byte, 0, len(d)+len(nid)), nid...), d...,
```
cristaloleg, line 192: Nit: looks like https://pkg.go.dev/slices#Concat

proof.go
```diff
@@ -141,6 +142,61 @@ func (proof Proof) IsEmptyProof() bool {
 	return proof.start == proof.end && len(proof.nodes) == 0 && len(proof.leafHash) == 0
 }
 
+func (proof Proof) isValidEmptyRangeProof(nth *NmtHasher, nID namespace.ID, root []byte, leaves [][]byte, checkNS bool) bool {
```
rootulp, line 145: 1. Could use unit tests
2. Could use GoDoc
3. What is an empty range proof?

proof.go
```diff
@@ -141,6 +142,61 @@ func (proof Proof) IsEmptyProof() bool {
...
+
+// ComputeAndValidateLeafHashes validates and hashes a list of leaves using the provided NMT hasher.
+func ComputeAndValidateLeafHashes(nth *NmtHasher, nid namespace.ID, leaves [][]byte) ([][]byte, error) {
+	hashes := make([][]byte, len(leaves))
+	for i, d := range leaves {
+		if nth.ValidateLeaf(d) != nil {
+			return nil, fmt.Errorf("leaf hash does not match the NMT hasher's hash format")
+		}
+		// check whether the namespace ID of the data matches the queried nID
+		if leadNid := namespace.ID(d[:nid.Size()]); !leadNid.Equal(nid) {
```
rootulp, line 173: should this be `leaf` and not `lead`?

proof.go
```diff
@@ -141,6 +142,61 @@ func (proof Proof) IsEmptyProof() bool {
...
+	// namespaces covered by the root 2) the root represents an empty tree, since
+	// it purports to cover the zero namespace but does not actually include
+	// any such nodes
+	return nID.Less(rootMin) || rootMax.Less(nID) || bytes.Equal(root, nth.EmptyRoot())
+}
+
+// ComputeAndValidateLeafHashes validates and hashes a list of leaves using the provided NMT hasher.
+func ComputeAndValidateLeafHashes(nth *NmtHasher, nid namespace.ID, leaves [][]byte) ([][]byte, error) {
+	hashes := make([][]byte, len(leaves))
+	for i, d := range leaves {
```
rootulp, line 168: [nit] rename `d` to `leaf`

proof.go
```diff
@@ -141,6 +142,61 @@ func (proof Proof) IsEmptyProof() bool {
...
+		if nth.ValidateLeaf(d) != nil {
+			return nil, fmt.Errorf("leaf hash does not match the NMT hasher's hash format")
```
rootulp, line 170: why does this error message contain "hash" when the leaf hasn't been hashed yet. I would've expected

```suggestion
			return nil, fmt.Errorf("leaf does not match the NMT hasher's hash format")
```

proof.go
```diff
@@ -141,6 +142,61 @@ func (proof Proof) IsEmptyProof() bool {
...
+		}
+		// check whether the namespace ID of the data matches the queried nID
+		if leadNid := namespace.ID(d[:nid.Size()]); !leadNid.Equal(nid) {
+			// conflicting namespace IDs in data
+			return nil, fmt.Errorf("leaf hash %x does not belong to namespace %x", d, nid)
```
rootulp, line 175: [nit] again why is hash used here b/c this leaf hasn't been hashed yet. I would've expected

```suggestion
			// conflicting namespace IDs in data
			return nil, fmt.Errorf("leaf %x does not belong to namespace %x", d, nid)
```

proof.go
```diff
@@ -141,6 +142,61 @@ func (proof Proof) IsEmptyProof() bool {
...
+		}
+		res, err := nth.HashLeaf(d)
+		if err != nil {
+			return nil, err
+		}
+		hashes[i] = res
```
rootulp, line 181: [nit]

```suggestion
		hash, err := nth.HashLeaf(d)
		if err != nil {
			return nil, err
		}
		hashes[i] = hash
```

proof.go
```diff
@@ -141,6 +142,61 @@ func (proof Proof) IsEmptyProof() bool {
...
+	}
+	return hashes, nil
+}
+
+// ComputePrefixedLeafHashes computes NMT leaf hashes for raw leaf data by prepending the given namespace ID.
+func ComputePrefixedLeafHashes(nth *NmtHasher, nid namespace.ID, leaves [][]byte) ([][]byte, error) {
+	hashes := make([][]byte, len(leaves))
+	for i, d := range leaves {
+		// prepend the namespace to the leaf data and hash it
+		res, err := nth.HashLeaf(slices.Concat(nid, d))
```
rootulp, line 191:
```suggestion
	for i, leaf := range leaves {
		// prepend the namespace to the leaf data and hash it
		res, err := nth.HashLeaf(slices.Concat(nid, leaf))
```

rootulp (commented):
Overall seems fine to me. Left a few optional nits and refactors that are safe to ignore if that code was just copied from the previous implementation.

It would be nice to include unit tests for the extracted helper methods.

proof.go
gupadhyaya: @rootulp this code was there before. i just pulled it out to a un-exported method for now. as @vgonkivs mentioned in his previous comment, empty proofs does not make sense. i will leave it here as is for now to make it backwards compatible and refactor this in future. will that work?

rootulp (approved):

proof.go
gupadhyaya: https://github.com/celestiaorg/nmt/issues/292

proof.go
gupadhyaya: have created an issue for coming back to this: https://github.com/celestiaorg/nmt/issues/292

walldiss (dismissed):
Nice changes and refactor! 

Agree with above that nmt pkg will benefit from unit tests for new exported funcs.

evan-forbes:
any remaining blocking reviews @vgonkivs ? or can we merge and flup

vgonkivs (approved):
LGTM

proof.go
```diff
@@ -320,30 +322,86 @@ func (proof Proof) VerifyLeafHashes(nth *NmtHasher, verifyCompleteness bool, nID
...
+// Parameters:
+//   - nth: The NMT hasher instance used for validation and root computation.
+//   - nID: The namespace ID that the proof is expected to correspond to.
+//   - leafHashes: The hashes of the leaves being proven.
+//   - isNamespace: A flag indicating whether namespace-specific validation should be performed.
+//
+// Returns:
+//   - The computed root hash if all checks pass.
+//   - An error if any validation fails or root computation fails.
+func (proof Proof) ComputeRootWithBasicValidation(nth *NmtHasher, nID namespace.ID, leafHashes [][]byte, isNamespace bool) ([]byte, error) {
```
rootulp, line 357: [blocking] this is a new exported function and it's lacking unit tests.

rootulp (changes requested):

proof.go
gupadhyaya: it doesn't add any new logic. Just a wrapper function using existing functions. Also, more tests will be covered as part of https://github.com/celestiaorg/nmt/issues/292

proof_test.go
```diff
@@ -1865,3 +1865,97 @@ func TestVerifySubtreeRootInclusion_infiniteRecursion(t *testing.T) {
...
+	leaves[0] = corruptedLeaf
+
+	leafHashes, err := ComputeAndValidateLeafHashes(hasher, nid, leaves)
+	require.Error(t, err) // Expect failure here
+
+	// Only call ValidateNamespace if hash step unexpectedly succeeds
+	if err == nil {
+		err = proof.ValidateNamespace(hasher, nid, leafHashes)
+		require.Error(t, err)
+	}
```
rootulp, line 1931: [blocking] this conditional will never evaluate to true because the test case will always err

```
	require.Error(t, err) // Expect failure here
```

This looks like AI generated slop.

```suggestion
```

rootulp (changes requested):

proof_test.go
gupadhyaya: fixed

vgonkivs (approved):

proof.go
```diff
@@ -248,64 +244,70 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, leaves [][]byt
...
-// The leafHashes parameter is a list of leaf hashes, where each leaf hash is represented
-// by a byte slice.
-// The size of leafHashes should match the proof range i.e., end-start.
-// If the verifyCompleteness parameter is set to true, the function also checks
-// the completeness of the proof by verifying that there is no leaf in the
-// tree represented by the root parameter that matches the namespace ID nID
-// outside the leafHashes list.
-func (proof Proof) VerifyLeafHashes(nth *NmtHasher, verifyCompleteness bool, nID namespace.ID, leafHashes [][]byte, root []byte) (bool, error) {
+// ValidateProofStructure checks ranges, leaf and node formats, and input compatibility.
+func (proof Proof) ValidateProofStructure(nth *NmtHasher, nID namespace.ID, leafHashes [][]byte) error {
```
rootulp, line 248: [question] why does this need to be exported? If it's exported then it should have unit tests. It doesn't seem used outside this file so proposal to un-export it.

[uber nit] it is repetitive to include `Proof` in the name of this method because the struct is already a proof. For example, a call site will look like: `proof.ValidateProofStructure`. Consider renaming to `ValidateStructure` instead.

proof.go
```diff
@@ -248,64 +244,70 @@ func (proof Proof) VerifyNamespace(h hash.Hash, nID namespace.ID, leaves [][]byt
...
-// If there is an issue during the proof verification e.g., a node does not conform to the namespace hash format, then a proper error is returned to indicate the root cause of the issue.
-// The leafHashes parameter is a list of leaf hashes, where each leaf hash is represented
-// by a byte slice.
-// The size of leafHashes should match the proof range i.e., end-start.
-// If the verifyCompleteness parameter is set to true, the function also checks
-// the completeness of the proof by verifying that there is no leaf in the
-// tree represented by the root parameter that matches the namespace ID nID
-// outside the leafHashes list.
-func (proof Proof) VerifyLeafHashes(nth *NmtHasher, verifyCompleteness bool, nID namespace.ID, leafHashes [][]byte, root []byte) (bool, error) {
+// ValidateProofStructure checks ranges, leaf and node formats, and input compatibility.
```
rootulp, line 247: What is input compatibility? This godoc could be clearer. For example

```suggestion
// ValidateProofStructure returns an error if the proof's range is invalid, or the format of leafHash is incorrect, or ...
```

rootulp (commented):
[no change needed] I don't understand the refactor anymore because so many lines were modified. One idea for a future refactor is to split the proof struct into multiple structs. For example: one struct for absence proof and one for inclusion proof. Then we can split this large file into two smaller files.

I don't see any issues besides the AI slop in the tests so _seems_ safe to merge after addressing that comment.

proof.go
gupadhyaya: ok. in the earlier version we were using these individual fuctions, hence was exported. now everything is done as part of `ComputeRootWithBasicValidation`, hence no longer needs to be exported.

proof.go
gupadhyaya: unexported

rootulp (approved):

vgonkivs (approved):

PR #325, opened 2026-05-22
fix: bound proof span in VerifySubtreeRootInclusion before allocating ranges

evan-forbes:
`VerifySubtreeRootInclusion` calls `ToLeafRanges` (allocating O((End-Start)/subtreeWidth) ranges from an attacker-controlled `proof.End()`) before the cheap `len(subtreeRoots)` count check; this fix rejects proofs whose span exceeds `len(subtreeRoots)*subtreeWidth` before allocating.

This is strictly hardening with very limited production accessibility: the only in-tree caller (celestia-node's `CommitmentProof.Verify`) is not network-reachable and its width derivation already curbs allocation; the exposure is external consumers that verify untrusted proofs with a fixed small width, so fixing because it is cheap and correct.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/proof.go
+++ b/proof.go
@@ -550,6 +550,34 @@ func (proof Proof) VerifySubtreeRootInclusion(nth *NmtHasher, subtreeRoots [][]b
 		}
 	}
 
+	// Bound the proof span before constructing the leaf ranges. ToLeafRanges
+	// allocates one LeafRange per (at most) subtreeWidth leaves of the span
+	// [proof.Start(), proof.End()), and proof.End() is attacker-controlled.
+	// Each provided subtree root can cover at most subtreeWidth leaves, so a
+	// proof whose span exceeds len(subtreeRoots)*subtreeWidth leaves can never
+	// satisfy the len(subtreeRoots) == len(ranges) check below. Reject such
+	// proofs here, before allocating, to avoid memory/CPU amplification from a
+	// large attacker-chosen span.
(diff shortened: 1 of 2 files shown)


proof.go
```diff
@@ -550,6 +550,34 @@ func (proof Proof) VerifySubtreeRootInclusion(nth *NmtHasher, subtreeRoots [][]b
...
+	// produce at least ceil(span/subtreeWidth) ranges. If that already exceeds
+	// len(subtreeRoots) the count check below would fail, so reject now. The
+	// comparison is written with division to avoid overflowing the int
+	// multiplication len(subtreeRoots)*subtreeWidth on attacker-chosen spans:
+	// ceil(span/subtreeWidth) > len(subtreeRoots).
+	quotient := span / subtreeWidth
+	remainder := span % subtreeWidth
+	if quotient > len(subtreeRoots) || (quotient == len(subtreeRoots) && remainder != 0) {
+		return false, fmt.Errorf("proof span %d exceeds the maximum number of leaves that %d subtree roots of width %d can cover", span, len(subtreeRoots), subtreeWidth)
+	}
```
rach-id: I feel the comments could be summarised to one-two lines maximum. This is too much documentation in the code

rach-id (dismissed):

rach-id (approved):

PR #335, opened 2026-08-20
test: cover nextSubtreeSize

### LINKED ISSUE #265: add unit tests for `nextSubtreeSize`
Is this func `nextSubtreeSize` unit tested? if not, I'd suggest adding unit tests for it in order to ensure its behaviour is aligned with what it is intended for.

_Originally posted by @staheri14 in https://github.com/celestiaorg/nmt/pull/260#discussion_r1653343583_

loglapa:
Closes #265

## Summary

- add direct table-driven coverage for nextSubtreeSize
- cover ranges limited by start alignment and by remaining length
- cover zero, unaligned, and larger aligned starts

## Testing

- go test ./...
- go vet ./...
- golangci-lint run --timeout 10m
- git diff --check

### DIFF
--- a/proof_test.go
+++ b/proof_test.go
@@ -1177,6 +1177,27 @@ func TestLargestPowerOfTwo(t *testing.T) {
 	}
 }
 
+func TestNextSubtreeSize(t *testing.T) {
+	tests := []struct {
+		name       string
+		start, end uint64
+		expected   int
+	}{
+		{name: "single leaf", start: 0, end: 1, expected: 1},
+		{name: "full range from zero", start: 0, end: 8, expected: 8},
+		{name: "unaligned start", start: 1, end: 9, expected: 1},
+		{name: "limited by start alignment", start: 4, end: 12, expected: 4},
+		{name: "limited by remaining range", start: 8, end: 12, expected: 4},
+		{name: "large aligned start", start: 16, end: 24, expected: 8},
+	}
+
+	for _, tt := range tests {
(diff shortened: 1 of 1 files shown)


rach-id (changes requested):
Use this prompt to include the relevant test cases:

  > Update PR celestiaorg/nmt#335 to strengthen TestNextSubtreeSize in proof_test.go.
  >
  > The current table only uses power-of-two range lengths, and every positive even start is itself a power of two. Add table cases that exercise:
  >
  > - Rounding a non-power-of-two remaining span down, such as (start: 0, end: 12) → 8 and (start: 8, end: 14) → 4.
  > - Alignment for a composite even start, such as (start: 12, end: 20) → 4.
  >
  > Keep the change test-only unless investigation reveals an actual implementation defect. Use descriptive case names consistent with the existing table.
  >
  > After editing, run:
  >
  > go test ./...
  > go vet ./...
  > git diff --check
  >
  > Review the final diff to ensure the new cases independently constrain both behaviors implemented by nextSubtreeSize: start alignment and rounding based on the remaining range.

loglapa:
Addressed the requested coverage in a0841be: added independent cases for non-power-of-two remaining spans at `(0, 12)` and `(8, 14)`, plus composite-even start alignment at `(12, 20)`. Validation passed: `go test ./...`, `go vet ./...`, and `git diff --check`.

rach-id (approved):

PR #336, opened 2026-08-20
refactor: extract subtree root computation

### LINKED ISSUE #261: extract `computeRoot` to a separate method
I think this was simpler because `computeRoot` also operates on a few vars inside of the scope of the wrapping method (e.g. `subtreeRoots`). I think this was an optimization wrt to the recursive calls. I would merge as is but if benchmarks and readability improve or stay the same, I would not mind Evan's suggestion in a followup PR.

_Originally posted by @liamsi in https://github.com/celestiaorg/nmt/pull/260#discussion_r1648010073_

loglapa:
Closes #261

## Summary

- extract subtree-root reconstruction from its recursive closure
- keep proof-node and subtree-root consumption explicit through shared cursors
- preserve validation, error wrapping, and hashing behavior

## Testing

- go test ./... -run 'TestVerifySubtreeRootInclusion|TestToLeafRanges|TestNextLeafRange' -count=1
- go test ./...
- go build ./...
- go vet ./...
- golangci-lint run --timeout 10m
- git diff --check


loglapa:
The fork workflows are still awaiting maintainer approval (Go, lint, buf, and proto generation). I attempted to approve each workflow run, but GitHub restricts that action to repository admins (`403 Must have admin rights`). A maintainer approval is the only remaining step before CI can run.

vgonkivs (commented):
Changes visually look correct to me. One thing before we procced: can we compare the two approaches with a benchmark, since the closure was originally kept inline as a possible optimization. Could you add benchstat before/after numbers for VerifySubtreeRootInclusion (old closure vs. extracted method) to the PR? I expect parity -  the recursion is dominated by HashNode/SHA-256 - but let's have the numbers

loglapa:
Added a reproducible `BenchmarkVerifySubtreeRootInclusion` in d3f396c and compared the parent implementation (inline closure) against this branch (extracted method) with 15 samples per version on Apple M5 / darwin-arm64:

```text
                              inline closure    extracted method    vs base
sec/op                        1.797µ ± 2%       1.807µ ± 1%        ~ (p=0.129 n=15)
B/op                          1.234Ki ± 0%      1.234Ki ± 0%       ~ (p=1.000 n=15)
allocs/op                     20.00 ± 0%        20.00 ± 0%         ~ (p=1.000 n=15)
```

Command: `go test -run "^$" -bench "^BenchmarkVerifySubtreeRootInclusion$" -benchmem -count=15 -benchtime=500ms`, compared with official `benchstat`. There is no statistically significant runtime difference and allocations are identical. `go test ./...`, `go vet ./...`, and `git diff --check` pass. A full race run still finds the pre-existing `Test_Root_RaceCondition` race; the same targeted race test fails on the parent implementation.

Issue #338, opened 2026-08-21
NamespacedMerkleTree.Root has a data race during concurrent cache initialization


loglapa:
### Reproduction

On current `main` (`e2177ee`):

```sh
go test -race -run '^Test_Root_RaceCondition$'
```

The existing test starts 200 goroutines that call `Root()` on the same tree. The race detector reports concurrent access to `rawRoot` in `nmt.go`:

- write at line 532 vs. read at line 522
- write at line 532 vs. read at line 535

The test fails with `race detected during execution of test`.

### Expected

Concurrent `Root()` calls used by the existing regression test complete without a data race and return the same root.

### Actual

Lazy cache initialization reads and writes `rawRoot` concurrently. Normal CI does not expose this because the test suite is run without the race detector.

### Impact

Concurrent callers can trigger unsynchronized access to the cached root. Besides failing race-enabled builds, this leaves behavior dependent on a data race under legitimate concurrent read-style use.

### Acceptance criteria

- `go test -race -run '^Test_Root_RaceCondition$'` passes.
- Existing root calculation and cache semantics remain intact.
- The concurrent regression coverage is retained or strengthened.
- The supported goroutine-safety contract is documented or clarified, in coordination with #39.

This report is narrower than #39: it records a concrete, reproducible race in the existing concurrent `Root()` test. It intentionally does not prescribe a locking or cache implementation.

loglapa:
I would like to work on this. I will first reproduce the race on current main, choose the smallest synchronization change that preserves lazy caching semantics, retain focused concurrent coverage, and run the full race-enabled test suite before opening a PR.

PR #339, opened 2026-08-21
docs: clarify ComputeSubtreeRoot range

### LINKED ISSUE #266: Document inner node reference
It's not clear here how the start and end could reference an inner node. Shouldn't it be leaves?

I would just say `[start, end)` as it's more succinct.

_Originally posted by @adlerjohn in https://github.com/celestiaorg/nmt/pull/260#discussion_r1679471210_

loglapa:
Closes #266.

Clarify that `ComputeSubtreeRoot` accepts an end-exclusive range of leaf indices, `[start, end)`. The previous wording implied that the indices themselves referenced an inner node, which was the ambiguity called out in the original review.

### Validation

- `go test ./...`
- `go test -race ./... -skip ^'Test_Root_RaceCondition$'` (the excluded test is the independently tracked race in #338)\n- `go vet ./...`\n- `golangci-lint run`\n- `docker run --rm -v "$PWD:/workspace" -w /workspace golang:1.27 go test ./...`

### DIFF
--- a/nmt.go
+++ b/nmt.go
@@ -704,9 +704,7 @@ func (n *NamespacedMerkleTree) updateMinMaxID(id namespace.ID) {
 	}
 }
 
-// ComputeSubtreeRoot takes a leaf range and returns the corresponding subtree root.
-// Also, it requires the start and end range to correctly reference an inner node.
-// The provided range, defined by start and end, is end-exclusive.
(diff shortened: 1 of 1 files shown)


chatton (approved):
LGTM thanks

